PR #2, opened 2026-05-15
feat: spec

ananas-block:
- [ ] detailed spec for all SPP instructions
- [ ] sync api definitions with docs
- [ ] add pocket rpc
- [ ] add squads policy program spec (the current file is outdated)
- [ ] SPP compliance check
- [ ] simplify everything


docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
+# Squads Zone Program
+
+The Squads Zone Program configures a zone on the Solana Privacy Protocol (TSPP), which provides private transfers in a single Solana transaction. The zone adds compliance features, a co-signer, and smart-account support.
+TODO: lookup requirements again and tick all boxes in the intro.
+
+For compliance, the zone program configures an auditor encryption key and verifies a zk proof in every transfer instruction. The zone zk proof shows that all balance updates encrypted to an auditor-readable key, so the auditor can decrypt and index all balances.
```
swenhelius:
```suggestion
For compliance, the zone program configures an auditor encryption key and verifies a zk proof in every transfer instruction. The zone zk proof shows that all balance updates were encrypted to an auditor-readable key, so the auditor can decrypt and index all balances.
```

docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
...
+| `Shielded Address` | `` | |
+
+
+## Architecture
+
+![Squads Zone Program architecture](diagrams/squads_policy_program.png)
+
+Source: [`diagrams/squads_policy_program.dot`](diagrams/squads_policy_program.dot). Regenerate with `just render-diagrams`.
+
+The squads program builds on top of the Solana Privacy Program (SPP). The backend (indexer, prover, relayer, and auditor) builds balances and proofs; the user signs; the squads program verifies the [zone proof](#zone-proof) and CPIs SPP. Execution is either synchronous ([transact](#transact)) or asynchronous with a [proposal](#asynchronous-transfers) user flow.
```
swenhelius: is the backend guaranteed to be vertically integrated in this case? if yes, might make sense to state explicitly

docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
...
+The squads policy program has full control over the features exposed from the SPP and balances of zone users. The SPP ensures correct state transitions, that all UTXOs are fully backed by spl tokens and only the owner of the balance can spend it.
```
swenhelius:
```suggestion
The squads policy program has full control over the features exposed from the SPP and balances of zone users. The SPP ensures correct state transitions, that all UTXOs are fully backed by SPL tokens and only the owner of the balance can spend it. 
```

docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
...
+
+For compliance, the zone program configures an auditor encryption key and verifies a zk proof in every transfer instruction. The zone zk proof shows that all balance updates encrypted to an auditor-readable key, so the auditor can decrypt and index all balances.
+
+For smart accounts, the program supports asynchronous execution and user accounts with shared encryption keys. Asynchronous execution uses a proposal buffer that a co-signer or relayer executes after approval. User accounts store encryption keys shared between the auditor and one or more smart account holders. The auditor key and user keys can be safely migrated unilaterally by proving the new shared key was encrypted correctly.
+
+Balances are stored in unspent transaction outputs (UTXOs). For a seamless account-like experience, the auditor consolidates fragmented UTXO balances so users can spend their full balance in one transfer. The squads zone depends on its backend and co-signer for liveness. To ensure that users can withdraw funds without the Squads backend, users can sync their wallets from RPC data alone, decrypt locally, and withdraw without a co-signer.
```
swenhelius: it says "the auditor consolidates [...]" can this be anyone besides Squads themselves?

further below in Line 99 it names it "Merge service consolidates [...]" -- could help to reference squads/auditor here to ensure reader doesn't think it's a separate thing

docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
...
+| 12 | create_proposal | 11 | Queue a deposit, withdrawal, or transfer for async execution. | — | ViewingKeyAccount | Proposal (create) | Proposer signs (smart account) |
+| 13 | cancel_proposal | 12 | Cancel a queued proposal before execution. | — | — | Proposal (close) | Proposer / owner signs |
+| 14 | execute_proposal | 13 | Relayer/co-signer settles an approved proposal with the proof. | ✓ | Proposal, ZoneConfig, sender + recipient ViewingKeyAccount | SPP trees (CPI), Proposal (close) | Co-signer / relayer signs |
+| 15 | execute_key_update | 14 | Backend settles an approved key update proposal with the key rotation proof. | — | KeyUpdateProposal, ZoneConfig | ViewingKeyAccount, KeyUpdateProposal (close) | Zone backend signs (proof) |
+| 16 | cancel_key_update | 15 | Cancel a queued key update proposal before execution and reclaim rent. | — | — | KeyUpdateProposal (close) | Proposer / owner signs (smart account) |
+
+#### transact
+
+Verifies the zone proof, then CPIs SPP `zone_transact`, which verifies the SPP proof and settles the UTXO state transition. One entrypoint for deposit, withdrawal, and transfer; `public_amount` selects the operation.
```
swenhelius:
```suggestion
Verifies the zone proof, then CPIs SPP `zone_transact`, which verifies the SPP proof and settles the UTXO state transition. It's a single instruction which handles deposit, withdrawal, and transfer; `public_amount` implicitly selects the operation per call.
```

docs/squads_policy_program.md
```diff
@@ -0,0 +1,611 @@
...
+    /// Output ciphertexts, zone serialization. Checked by the zone proof, not parsed by SPP.
+    encrypted_utxos: Vec<u8>,
+}
+```
+
+**Encrypted UTXO Serialization**
+
+The `sender_viewing_key_account` and `recipient_viewing_key_account` identify the owners and serve as view tags, so the ciphertext contains no pubkeys or tags. A transfer moves one asset with no separate SOL change, so the sender has one change output and each recipient one. The asset stays private; each ciphertext includes its own `asset_id`.
+
+The sender change derives its blinding from the `blinding_seed` in its viewing key account as `Poseidon(blinding_seed, blinding_nonce)`, so only its amount and asset are transmitted, in a separate ciphertext encrypted to the sender's shared viewing key. Each recipient output transmits amount, asset, and the sender-chosen blinding, encrypted to the recipient's shared viewing key. One ephemeral `tx_viewing_pk` is shared across all ciphertexts.
```
swenhelius:
```suggestion
The sender change UTXO derives its blinding from the `blinding_seed` in its viewing key account as `Poseidon(blinding_seed, blinding_nonce)`, so only its amount and asset are transmitted, in a separate ciphertext encrypted to the sender's shared viewing key. Each recipient output transmits amount, asset, and the sender-chosen blinding, encrypted to the recipient's shared viewing key. One ephemeral `tx_viewing_pk` is shared across all ciphertexts.
```

docs/squads_policy_program.md
ananas-block: yes the squads backend will be fully vertically integrated.

docs/squads_policy_program.md
ananas-block: now outdated

docs/detailed_user_flows.md
```diff
@@ -0,0 +1,256 @@
...
+    participant Client as Client<br>(Wallet + Swaps)
+    participant PocketRPC as Pocket RPC<br>(Photon / Prover / Relayer)
+    participant Policy as Policy Program
+    participant System as System Program<br>(Shielded Pool)
+    participant Trees as Tree accounts
+
+    Note over Client: Build transaction
+    Client->>PocketRPC: get_balance
+    PocketRPC-->>Client: balance
+    Note over Client: 1. Set amount <br> 2. set recipient address (in) <br> 4. sign recipient address and amount
```
sergeytimoshin, line 209: - what is "(in)" in recipient address? 
- 3 point is missing

docs/detailed_user_flows.md
```diff
@@ -0,0 +1,256 @@
...
+	  Client->>PocketRPC: send transaction <br>(recipient, amount, signature)
+    PocketRPC-->>Policy: submit tx<br>policy_transact
```
sergeytimoshin, line 211: > "send transaction" ... => "submit tx" 

stylistically, it’s better to use the same abbrev everywhere or not use it at all

docs/spec.md
```diff
@@ -0,0 +1,1882 @@
...
+    /// `KDF(ECDH(signing_sk, sync_pk)) · G`.
+    viewing_pk: P256Pubkey,
+    /// Unix seconds; set at the moment the entry is appended.
+    created_at: i64,
+}
+```
+
+Invariants:
+
+- The current delegate is set if and only if `entries` is non-empty.
```
swenhelius, line 1779: @ananas-block is this "iff" on purpose? "if and only if". if yes, then both must be true at the same time.

but as per the "revoke" instruction, when we delete the delegate, we still keep the entries untouched, which contradicts the above

ananas-block: yes iff defines only the following two states as valid right?
1. delegate set and at least one entry
2. delegate not set and no entries

PR #3, opened 2026-05-19
chore: use helius-labs/privacy-program-libs

sergeytimoshin:


ananas-block:
**Confirmed: `PoolTreeMutationFailed` is overloaded across three semantically distinct failure surfaces.**

| Path | Site | What actually went wrong |
|---|---|---|
| `append_state_leaves/processor.rs:19` | `append_to_pool` | `SparseMerkleTree::append` / state next_index advance failed |
| `insert_addresses/processor.rs:32` | `insert_address_into_queue` | Address queue full, bloom filter saturation, or duplicate |
| `batch_update_address_tree/processor.rs:39` | `update_tree_from_address_queue` | **Groth16 proof verification failed**, batch not ready, or root mismatch |

These are not related failure modes — the batch-update case is the cryptographic-proof rejection path, which is the single most diagnostically important error in this whole program. Collapsing it to the same code as "state append index overflow" means an on-chain log of `PoolTreeMutationFailed` tells you nothing about which subsystem actually rejected the transaction.

Worth noting `InvalidPoolTreeAccounts` is similarly overloaded — 5 sites across the three processors (`batch_update_address_tree/processor.rs:25,27`, `insert_addresses/processor.rs:26,28`, plus loader rejections). But that one's more defensible since it's all "couldn't open the account as expected layout." The mutation-side collapse is the painful one.

Suggested split:
- `StateAppendFailed` — state sub-tree append errors
- `AddressQueueInsertFailed` — queue/bloom errors
- `BatchProofVerificationFailed` — Groth16 / batch update errors

ananas-block (approved):

ananas-block:
probably makes sense to propagate the errors else lgtm

PR #4, opened 2026-05-19
chore: externalize Light fixtures to GitHub release

sergeytimoshin:
## Summary
- Adds `tools/build-fixtures.sh` to bundle `sdk-libs/cli/{bin,accounts}` plus a `MANIFEST.json` into `light-fixtures.tar.gz`.
- Adds `.fixtures-version` (`light-fixtures-zkc-0.28.4`) as the single source of truth for the pinned release tag.
- `just fetch-fixtures` downloads + `shasum -c`s into `${ZOLANA_CACHE_DIR:-$HOME/.cache/zolana}/fixtures/<tag>/`, idempotent on warm cache.
- `build-light-programs` now depends on `fetch-fixtures` and cps the .so files from the cache.
- `sdk-libs/cli/src/main.rs` resolves bins and accounts from the cache dir after env override / `target/deploy`.
- CI (`setup-forester-e2e`) caches `~/.cache/zolana/fixtures` keyed on the pinned tag and points `LIGHT_PROTOCOL_ACCOUNTS_DIR` at the cache.

The v1 release `light-fixtures-zkc-0.28.4` is published at https://github.com/helius-labs/zolana/releases/tag/light-fixtures-zkc-0.28.4 with the bundle byte-identical to what's currently in `sdk-libs/cli/{bin,accounts}`.

`sdk-libs/cli/{bin,accounts}` is **not** removed in this PR — a follow-up commit (or this PR before merge) will `git rm` them once we're confident the fetch path works end-to-end in CI.

Stacked on #3 (base: `chore/ext-program-libs`).

## Test plan
- [ ] CI green on this branch
- [ ] Locally: `rm -rf ~/.cache/zolana/fixtures && just fetch-fixtures` downloads + verifies
- [ ] Locally: `just build-light-programs` copies bins from cache into `target/deploy`
- [ ] Worktree fixtures removed in a follow-up commit (or before merge) once green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8, opened 2026-05-20
feat(shielded-pool): tighten forester auth + event logs + e2e tests

sergeytimoshin:
Stacks on top of #3.

## Summary
- **Auth**: `batch_update_address_tree` now rejects any signer that isn't the registry's CPI authority PDA (`UnauthorizedCaller`, `Custom(6)`). `cpi_authority_bump` rides on `BatchUpdateAddressTreeData`; new `programs/shielded-pool/src/pda.rs` derives via `sol_sha256` so we don't need to bump pinocchio to 0.11 just to pull in `pinocchio-pubkey`.
- **Events**: wire `light-event` into `shielded-pool` and emit `PoolTreeCreated`, `StateLeavesAppended`, `AddressQueued`, `AddressTreeBatchUpdated` from each processor via `sol_log_data`.
- **Integration tests**: litesvm coverage for all four shielded-pool instructions plus a full registry chain (`initialize_protocol_config` → `register_forester` → `register_forester_epoch` → warp → `finalize_registration` → `forest_address_tree`) that proves the CPI authority survives shielded-pool's auth check. New minimal `registry_sdk` helper in `sdk-libs/program-test`, modeled on light-protocol's. `justfile` gains a `test-litesvm` target.
- **Dep cleanup**: absorb privacy-program-libs PR #4 — drop `light-indexed-merkle-tree` (v1 non-inclusion helper now delegates to `light-merkle-tree-reference::indexed`) and `light-zero-copy`. All workspace deps track upstream `main` again.

## Test plan
- [x] `cargo test --workspace --exclude light-prover-client` (excludes the pre-existing prover-server test that needs the `light-prover` binary)
- [x] `cargo clippy --workspace --tests --exclude light-prover-client` — only pre-existing `pool_tree_buffer.rs` doc warning remains
- [x] `cargo build-sbf -p light-registry`
- [x] `cargo build-sbf --manifest-path programs/shielded-pool/Cargo.toml -- --features bpf-entrypoint`
- [x] `cargo test -p light-program-test` (4 end-to-end + 2 registry-CPI tests pass against the loaded `.so`s)


ananas-block:
> don't need to bump pinocchio to 0.11 just to pull in pinocchio-pubkey.

should we bump to pinocchio 0.11 ?

ananas-block:
Auth flow is more complex than it needs to be — please simplify by hardcoding the registry's canonical CPI authority PDA as a `[u8; 32]` constant in `zolana-interface` and reducing `verify` to a single equality check against `signer.address()`.

That lets us drop:
- `cpi_authority_bump` from `BatchUpdateAddressTreeData`
- `programs/shielded-pool/src/pda.rs` (and the `derive_pda` sha256 path)
- the `sha2` host-fallback dep on `shielded-pool`
- the `bump` plumbing through `forest_address_tree_once` mutating `params`

The registry already calls `find_program_address` to sign; the receiver doesn't need the bump at all. Add a unit test asserting the hardcoded constant matches `Pubkey::find_program_address(&[b"cpi_authority"], &LIGHT_REGISTRY_PROGRAM_ID)` so an upstream rename of the seed/program-id is loud.

ananas-block:
Let's drop the `sol_log_data`-based events. Transaction logs get truncated when CU/log thresholds are hit, so any indexer relying on them is fragile — especially for `StateLeavesAppended` where the payload scales with `leaves.len()`.

Index the instructions directly instead — drop the emit code entirely. Indexers already see the outer `BatchUpdateAddressTree` / `AppendStateLeaves` / `InsertAddresses` / `CreatePoolTree` ix data on the wire, and can reconstruct derived fields (`new_root`, `start_index`, `queue_next_index`) from pre/post account data deltas. Zero on-chain cost, no extra event types to keep in sync with the program, no log-truncation risk.

PR #11, opened 2026-05-28
Sergey/spp zk

sergeytimoshin:


sergeytimoshin:
Split into #13 for circuits/tests and #14 for the surrounding program/CLI wiring.

PR #12, opened 2026-05-28
fix p256 => bn254 conversion

sergeytimoshin:
P256/Solana pubkeys are raw 32-byte values and cannot be passed directly into BN254 Poseidon. This updates the spec to use canonical hash-to-field encoding for opaque keys, binds P256 signatures to the derived owner key field, and requires nullifier_pk to be a canonical BN254 field element.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -318,8 +318,6 @@ Symmetric key to derive nullifiers.
 
 `nullifier_pk := Poseidon(nullifier_secret)`
 
-`nullifier_secret` is a 31-byte big-endian value; at 31 bytes it is always `< Fr`, so it is a valid field element by construction. `nullifier_pk`, a 32-byte Poseidon output, must be serialized canonical BN254 (`< Fr`).
-
 **Methods:**
 
 - `nullifier_pk() -> [u8; 32]` — returns `nullifier_pk` (defined above).
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -266,22 +266,28 @@ In compressed form the signing and nullifier public keys are compressed in an ow
 
 `CompressedShieldedAddress = (owner_hash, viewing_pk)`
 
-Raw public keys are encoded as 128-bit limbs and absorbed with 2-input Poseidon.
+## Pubkey Field Encoding
+
+A pubkey is encoded into a single BN254 field via nested 2-input Poseidon over 128-bit big-endian limbs. This `pk_field(pk)` is the canonical form used wherever a pubkey appears inside a Poseidon hash anywhere in this spec.
 
 ```
-P256:
-x_hash        = Poseidon(x_low_128, x_high_128)
-parity_x_hash = Poseidon(y_is_odd, x_hash)
-owner_hash    = Poseidon(parity_x_hash, nullifier_pk)
+P256 (33 B SEC1: parity || x_be32):
+  x_hash := Poseidon(x_low_128, x_high_128)
+  pk_field(pk) := Poseidon(y_is_odd, x_hash)
 
-Solana / Ed25519:
-pk_hash    = Poseidon(pk_low_128, pk_high_128)
-owner_hash = Poseidon(pk_hash, nullifier_pk)
+Solana / Ed25519 (32 B):
+  pk_field(pk) := Poseidon(pk_low_128, pk_high_128)
 ```
 
-SPP recomputes the P256 path from the witnessed point and the Solana path from the signer account.
+P256's extra `y_is_odd` layer means the two encodings cannot collide except via Poseidon preimage collision (~2⁻²⁵⁴); no explicit scheme tag is needed.
 
-Wherever a P256 pubkey (e.g. `user_viewing_pk`, `merge_authority_pubkey`) appears as an input to a Poseidon hash elsewhere in this spec, it is absorbed via the same chain: `Poseidon(y_is_odd, Poseidon(x_low_128, x_high_128))`.
+## Owner Hash
+
+```
+owner_hash := Poseidon(pk_field(signing_pk), nullifier_pk)
+```
+
+The proof recomputes `pk_field(signing_pk)` from a witnessed P256 point; for Solana / Ed25519 inputs, SPP supplies it from the verified signer account.
 
 # Signing Key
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -443,7 +443,44 @@ In compressed form the signing and nullifier public keys are compressed in an ow
...
+All Poseidon inputs are canonical BN254 field elements (`< Fr`).
+
+Opaque bytes are mapped with the Light-style convention:
+
+```
+HashToFieldBE(domain, fixed_width_parts...) :=
+    digest = Keccak256(domain || fixed_width_parts... || 0xff)
+    digest[0] = 0
+    digest as 32-byte big-endian field element
+```
```
ananas-block: Keccak256 is too expensive in the circuit

docs/spec.md
```diff
@@ -443,7 +443,44 @@ In compressed form the signing and nullifier public keys are compressed in an ow
...
+
+Variable-length inputs MUST be length-delimited before `HashToFieldBE`. Setting the high byte to zero is a truncation-to-field convention, not modular reduction.
+
+```
+owner_hash = Poseidon(OWNER_HASH_TAG, owner_scheme, owner_key_field, nullifier_pk)
+```
+
+`OWNER_SCHEME_P256 = 1` for P256 owners; `OWNER_SCHEME_SOLANA = 2` for Solana / Ed25519 owners. `OWNER_HASH_TAG` and `P256_OWNER_KEY_TAG` are fixed domain-separation field constants (concrete values TBD; tracked as a protocol-constants table). They MUST be distinct from each other and from all `DOM_SEP_*`, `ENABLE_TAG`, and `REVOKE_TAG` constants.
```
ananas-block: so far I avoided tags like this because adding an additional input for a poseidon hash is considerable overhead

docs/spec.md
```diff
@@ -443,7 +443,44 @@ In compressed form the signing and nullifier public keys are compressed in an ow
 
 `CompressedShieldedAddress = (owner_hash, viewing_pk)`
 
-`owner_hash = Poseidon(signing_pk_a, signing_pk_b, nullifier_pk)`
```
ananas-block, line 269: What if we just split it 4 ways?
`owner_hash = Poseidon(signing_pk_a_low,signing_pk_a_high,signing_pk_b_low,signing_pk_b_high, nullifier_pk)`

ananas-block: or poseidon hash each coordinate into Poseidon(signing_pk_a_low,signing_pk_a_high) and hash that hash into the owner hash

docs/spec.md
sergeytimoshin: dropped

docs/spec.md
sergeytimoshin: dropped

docs/spec.md
sergeytimoshin: updated:

```
P256 (compressed: parity + x):
  x_hash        = Poseidon(x_low_128, x_high_128)
  parity_x_hash = Poseidon(y_is_odd, x_hash)
  owner_hash    = Poseidon(parity_x_hash, nullifier_pk)

Solana / Ed25519:
  pk_hash       = Poseidon(pk_low_128, pk_high_128)
  owner_hash    = Poseidon(pk_hash, nullifier_pk)
```

docs/spec.md
```diff
@@ -1446,25 +1478,24 @@ struct EnableMergeAuthorityIxData {
...
     /// Lets the wallet rediscover its own enables on restore-from-mnemonic
     /// by scanning the same bootstrap stream it scans for incoming transfers.
     recipient_bootstrap_view_tag: [u8; 32],
     /// AES-256-GCM ciphertext over `MergeAuthorityNotification`, encrypted
     /// to `user_viewing_pk` using the shared `tx_viewing_pk` and
     /// `info = "zolana/merge_authority_notify_bootstrap"`. Same plaintext layout
-    /// as `ciphertext_authority`. 74 B plaintext + 16 B GCM tag.
-    ciphertext_bootstrap: [u8; 90],
+    /// as `ciphertext_authority`. 108 B plaintext + 16 B GCM tag.
+    ciphertext_bootstrap: [u8; 124],
```
ananas-block: why are the ciphertexts bigger now?
it should be still sufficient to store the compressed pubkey, the cipher text sizes shouldnt change unless they were wrong to begin with

docs/spec.md
```diff
@@ -443,7 +443,34 @@ In compressed form the signing and nullifier public keys are compressed in an ow
...
+```
+pk_low     = u128_be(solana_pubkey[0..16])
+pk_high    = u128_be(solana_pubkey[16..32])
+pk_hash    = Poseidon(pk_low, pk_high)
+owner_hash = Poseidon(pk_hash, nullifier_pk)
+```
+
+SPP derives `(pk_low, pk_high)` from the signer account; the proof receives `pk_hash` as a public input bound into `owner_hash`.
+
+Scheme separation: a P256 owner's chain folds in an extra `parity_x_hash` layer that Solana's chain does not, so the two encodings cannot collide except via a Poseidon preimage collision (~2⁻²⁵⁴). No explicit scheme tag is needed.
```
ananas-block: this is still too much verbose and slop

docs/spec.md
```diff
@@ -470,6 +497,8 @@ Symmetric key to derive nullifiers.
 
 `nullifier_pk := Poseidon(nullifier_secret)`
 
+`nullifier_secret` is a 31-byte big-endian integer and is therefore always `< Fr` (since 2²⁴⁸ < Fr). Serialized `nullifier_pk` MUST decode to a canonical BN254 field element (`< Fr`); checking only the top bits is insufficient.
```
ananas-block: also here too much ai description

docs/spec.md
```diff
@@ -293,6 +314,8 @@ Symmetric key to derive nullifiers.
 
 `nullifier_pk := Poseidon(nullifier_secret)`
 
+`nullifier_secret` is a 31-byte big-endian field element. Serialized `nullifier_pk` must be canonical BN254 (`< Fr`).
```
ananas-block:
```suggestion
```
Its 31 bytes it cannot be larger than BN254 Fr size

docs/spec.md
```diff
@@ -792,14 +814,16 @@ external_data_hash := Sha256BE(
 )
 ```
 
+`Sha256BE` in field context means SHA-256 over the byte preimage, then `digest[0] = 0`, interpreted as a BN254 field element.
```
ananas-block: we should define this in the Glossar

PR #15, opened 2026-05-29
add external_data_hash to SPP and Merge proof public inputs

sergeytimoshin:


ananas-block (commented):
makes sense can you rewrite the description in non ai slop pls,
eg dispatched instruction, or folded into ...

PR #16, opened 2026-05-29
document signed public-amount encoding

sergeytimoshin:


docs/spec.md
```diff
@@ -764,9 +764,12 @@ struct MergeEncryptedUtxo {
...
+| public_sol_amount | instruction data; raw unsigned `u64` magnitude of the public SOL movement. |
+| public_spl_amount | instruction data; raw unsigned `u64` magnitude of the public SPL movement. |
 | public_spl_asset_pubkey | derived by SPP from the vault token account's mint |
+| public_amount_mode | instruction data; `0` = transfer, `1` = shield, `2` = unshield. Selects the sign the proof applies to `public_sol_amount` / `public_spl_amount` in balance conservation; constrained in-circuit to `{0, 1, 2}`. |
+| relayer_fee | instruction data (`u16`); SOL paid to the relayer. Subtracted from the shielded SOL balance unconditionally (independent of `public_amount_mode`), since the fee is owed on any relayer-submitted transaction — including a plain transfer. Also committed in `external_data_hash`. |
```
ananas-block: we can just add the relayer fee on top of the public sol amount it does not need to be a separate circuit input

docs/spec.md
```diff
@@ -764,9 +764,12 @@ struct MergeEncryptedUtxo {
 | utxo_tree_roots (one per input UTXO) | resolved from `utxo_tree_root_index[i]` against the root cache of the input's UTXO tree |
 | nullifier_tree_roots (one per input UTXO) | resolved from `nullifier_tree_root_index[i]` against the root cache of the input's nullifier tree |
 | private_tx_hash | instruction data |
-| public_sol_amount | instruction data |
-| public_spl_amount | instruction data |
+| external_data_hash | instruction data; SPP recomputes it from the dispatched instruction (see the **external_data_hash** definition below) and checks it equals this public input. It is a standalone public input — not only folded into `private_tx_hash` — because SPP cannot recompute `private_tx_hash` (the input UTXO hashes it commits to are private), so binding the proof to the dispatched instruction requires `external_data_hash` to be exposed directly. |
+| public_sol_amount | instruction data; raw unsigned `u64` magnitude of the public SOL movement. |
+| public_spl_amount | instruction data; raw unsigned `u64` magnitude of the public SPL movement. |
```
ananas-block: we could use the same trick that tornado nova uses [see github](https://github.com/tornadocash/tornado-nova/blob/f9264eeffe48bf5e04e19d8086ee6ec58cdf0d9e/circuits/transaction.circom#L122)
the these are 32 byte values not u64s and we subtract by using modular arithmetic, in instruction data we do use u64 and a marker whether it is shield or unshield and produce the 32 byte value onchain

docs/spec.md
ananas-block: every public circuit input costs us 800 CU

docs/spec.md
```diff
@@ -1622,7 +1628,7 @@ struct GenerateSppProofResponse {
 
 ## Relayer
 
-Signs and submits a Solana transaction on behalf of a user, pays the SOL transaction fee on the Solana payer slot, and is reimbursed plus rewarded out of the `relayer_fee` field included in the shielded instruction (see [`transact`](#transact)). The relayer cannot change the user's shielded transactions: the SPP proof commits to all transaction parameters. The relayer never sees plaintext UTXOs; it only signs as the Solana payer.
+Signs and submits a Solana transaction on behalf of a user, pays the SOL transaction fee on the Solana payer slot, and is reimbursed plus rewarded out of the `relayer_fee` field included in the shielded instruction (see [`transact`](#transact)). This holds for shielded transfers as well as unshields: a transfer is `public_amount_mode = 0` with `relayer_fee > 0`, and the proof debits `relayer_fee` from the sender's shielded SOL balance regardless of mode ([`transact`](#transact) check 8), so the fee need not be encoded as a withdrawal. The relayer cannot change the user's shielded transactions: the SPP proof commits to all transaction parameters. The relayer never sees plaintext UTXOs; it only signs as the Solana payer.
```
ananas-block: is this necessary?

PR #14, opened 2026-05-28
Wire SPP programs and CLI flows

sergeytimoshin:
## Summary
- wire SPP instructions, generated verifying keys, nullifier tree update handler, and program tests
- add prover CLI setup/prove/export commands and fixture generation
- add pocket CLI shielded wallet/SOL flows, demo script, and E2E harness plumbing

Stacked on #13.

## Test plan
- GOCACHE=/private/tmp/zolana-go-cache ../../../light/.local/go/bin/go test ./prover/poseidon ./prover/spp
- bash -n scripts/pocket-sol-demo.sh
- cargo check -p zolana-cli --tests
- cargo test -p zolana-cli --test pocket_cli_e2e pocket_cli_creates_spec_p256_wallet

PR #13, opened 2026-05-28
Add SPP circuits and prover tests

sergeytimoshin:
## Summary
- add BN254 Poseidon helpers and golden/circuit tests
- add SPP transaction circuit shapes, P256 owner checks, balance checks, and public input hashing
- add SPP nullifier tree update circuit and prover-side tests

## Test plan
- GOCACHE=/private/tmp/zolana-go-cache ../../../light/.local/go/bin/go test ./prover/poseidon ./prover/spp

PR #26, opened 2026-06-03
Add SPP interface wire format

sergeytimoshin:
Split from #14.

Adds SPP interface instruction/config wire data and roundtrip tests. Treats encrypted_utxos as opaque instruction bytes; encryption policy is outside this PR.

Verification:
- cargo test -p zolana-interface


sergeytimoshin:
This PR was auto-marked merged while restacking because its old base branch (#13) was rewritten to include the interface commit. Replacement review PR is #29, with the same interface branch retargeted to the correct base.

PR #27, opened 2026-06-03
Add SPP prover tooling

sergeytimoshin:
Split from #14.

Adds SPP prover-server CLI commands, BSB22 proof encoding support, xtask verifying-key generation, and justfile SPP key-generation targets.

Verification:
- cargo check --offline -p xtask
- go test ./... in prover/server

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7570,6 +7570,8 @@ dependencies = [
 name = "xtask"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


sergeytimoshin:
Superseded by #34 — the SPP stack was rebased onto a fresh `main` and re-pushed under clean branch names (`spp/1-circuit … spp/7-demo`). Continuing review on the new stack.

PR #24, opened 2026-06-03
Add pocket CLI SPP commands and e2e

sergeytimoshin:
Restacked after #25. Adds pocket CLI commands for wallet creation, pool tree initialization, shield, transfer, unshield, balance queries, proof-bundle transaction submission, the local pocket SOL demo script formerly in #28, and CLI e2e coverage.

Verification run on the final stack:
- cargo test --offline -p zolana-interface
- cargo check --offline -p shielded-pool-program --lib --tests
- cargo test --offline -p shielded-pool-program --lib --tests
- cargo check --offline -p zolana-cli --tests
- go test -count=1 ./... in prover/server

PR #28, opened 2026-06-03
Add pocket SOL demo script

sergeytimoshin:
Split from #14.

Adds a local solana-test-validator demo script for the pocket SOL shield, private transfer, P256, and unshield flows. This depends on the pocket CLI and prover key-generation targets from the previous PRs.

Verification:
- not run end-to-end; script is separated from compile-tested CLI/e2e coverage


sergeytimoshin:
Folded into #24 while restacking: the demo script now lives in the CLI PR alongside the commands and CLI e2e coverage.

PR #32, opened 2026-06-10
feat(spp): add proofless shield with spec-aligned program cleanup

sergeytimoshin:


fixtures/vendor/spl-noop.lock
ananas-block, line 1: do we need tis?

sergeytimoshin:
Superseded by the split stack: #46 for core shielded-pool/spec alignment and #47 for CLI/localnet/fixtures/CI tooling.

PR #41, opened 2026-06-11
spec: global shared viewing key

ananas-block:
Summary:
1. proof of concept for global shared viewing key distribution
2. global shared viewing key spec.md that specifies how to adjust the main spec.md


ananas-block:
**Possible mvp:**
- only use one relayer (co-signer), cosigner decrypts data locally
- this way we dont need verifiable encryption, or user accounts etc.

PR #43, opened 2026-06-12
add user registry program

swenhelius:
adds pinocchio user-registry program in line with spec

the program maps owner Solana pubkey to static shielded keys + optional sync delegate 

it creates 1 PDA account per user, and it has 4 instructions (register, set_sync_delegate, rotate_sync_delegate_key, revoke_sync_delegate)

CHECKS: P-256 prefix checks, canonical nullifier (< Fr); we do not verify correct KDF  onchain

we also updated docs/spec.md to be in sync with the functions names etc


programs/user-registry/src/log.rs
ananas-block, line 1: interesting

ananas-block: we can also use solana_msg::msg! althought I didn't need logs for the my last program at all

programs/user-registry/src/state.rs
```diff
@@ -0,0 +1,145 @@
...
+            std::cmp::Ordering::Less => return true,
+            std::cmp::Ordering::Greater => return false,
+            std::cmp::Ordering::Equal => {}
+        }
+    }
+    false
+}
+
+#[cfg(test)]
+mod tests {
```
ananas-block: to keep audit scope minimal I usually put all tests into a tests dir

programs/user-registry/src/log.rs
swenhelius: kk switched and removed the log.rs

note that the shielded pool program on main has a log.rs file too

program-libs/interface/src/user_registry/mod.rs
```diff
@@ -0,0 +1,24 @@
+pub mod instruction;
+pub mod state;
+
+pub use state::{SyncDelegateEntry, UserRecord};
+
+pub const USER_REGISTRY_PROGRAM_ID: [u8; 32] = [
+    122, 111, 108, 97, 110, 97, 95, 117, 114, 101, 103, 95, 118, 48, 95, 95, 95, 95, 95, 95, 95,
+    95, 95, 95, 95, 95, 95, 95, 95, 95, 95, 1,
+];
```
ananas-block: can we use a pubkey_array!("") macro?

programs/user-registry/src/validation.rs
ananas-block, line 1: this validation is pretty weak, it doesnt add security, I think we can remove it.

swenhelius: removed

ananas-block (approved):

PR #45, opened 2026-06-13
refactor(prover): reorganize spp prover into circuits/prover/prover-test

ananas-block:
Restructure the Go prover server: move circuit sources under circuits/, split the SPP transaction prover into prover/nullifier_tree, and relocate tests and protocol helpers under prover-test/. Remove the legacy v1/v2 prover packages and the old standalone poseidon package.

1. add client crate
2. remove prover client crate it is now part of client
3. add Transaction abstraction in client crate with integration tests that use prover server with rust inputs


prover/server/circuits/spp_transaction/p256.go
```diff
@@ -1,13 +1,14 @@
 package transaction
 
 import (
-	"light/light-prover/prover/poseidon"
+	gadgetlib "light/light-prover/circuits/gadget"
```
sergeytimoshin: prob makes sense to rename package names too
`light/light-prover` => `zolana/prover`

sdk-libs/client/src/transaction.rs
```diff
@@ -0,0 +1,529 @@
...
+                    utxo_hash,
+                    nullifier,
+                })
+            })
+            .collect()
+    }
+
+    pub fn into_prover(
+        self,
+        resolver: &mut impl ProofResolver,
```
ananas-block: reanme trait ProofResolver to Rpc, add proper rpc calls similar to dev/light-protocol/sdk-libs/client Rpc trait

ananas-block: resolve  should be renamed to get_input_utxo_merkle_proofs

sdk-libs/client/src/transaction.rs
```diff
@@ -0,0 +1,529 @@
...
+    fn asset_id(&self, assets: &AssetRegistry, asset: &Address) -> Result<u64, ClientError> {
+        if asset == &SOL_MINT {
+            Ok(SOL_ASSET_ID)
+        } else {
+            Ok(assets.asset_id(asset)?)
+        }
+    }
+}
+
+pub struct SignedTransaction {
```
ananas-block: move into separate file

prover/server/prover/common/marshal.go
```diff
@@ -315,7 +315,7 @@ func (ps *TransferProofSystem) UnsafeReadFrom(r io.Reader) (int64, error) {
 func ReadSystemFromFile(path string) (interface{}, error) {
 	if strings.Contains(strings.ToLower(path), "transfer-eddsa") {
```
sergeytimoshin: it looks fragile, and its already broken after transfer_eddsa_2_3 => trasfer rename.

maybe we infer circuit type from the serialized key itself, like we do here:
1. https://github.com/helius-labs/zolana/blob/main/prover/server/prover/spp/prover/transaction/proving_system.go#L188-L206
2. and then https://github.com/helius-labs/zolana/blob/main/prover/server/prover/spp/prover/transaction/proving_system.go#L38-L42

prover/server/server/server.go
sergeytimoshin, line 1098: delete

sergeytimoshin: delete

prover/server/prover-test/spp/prover/transaction/field_derivation_test.go
sergeytimoshin, line 120: this file doesn't exist anymore, test will fail

prover/server/prover-test/poseidon/poseidon.go
```diff
@@ -0,0 +1,55 @@
...
+// HashWithT computes Poseidon(domain_tag = 0, inputs...) at width t; requires
+// len(inputs) == t-1.
+func HashWithT(t int, inputs []*big.Int) (*big.Int, error) {
+	if t < MinWidth || t > MaxWidth {
+		return nil, fmt.Errorf("poseidon: unsupported width t=%d", t)
+	}
+	if len(inputs) != t-1 {
+		return nil, fmt.Errorf("poseidon: want %d inputs for t=%d, got %d", t-1, t, len(inputs))
+	}
+	return iden3.Hash(inputs)
```
sergeytimoshin, line 47:
```suggestion
	for i, input := range inputs {
		if err := ValidateField(fmt.Sprintf("input[%d]", i), input); err != nil {
			return nil, err
		}
	}

	defer func() {
		if r := recover(); r != nil {
			out = nil
			err = fmt.Errorf("poseidon: hash panic: %v", r)
		}
	}()
	return iden3.Hash(inputs)
```

prover/server/main.go
```diff
@@ -143,12 +88,63 @@ func runCli() {
 					return nil
 				},
 			},
+			{
+				Name: "setup-transfer",
+				Flags: []cli.Flag{
+					&cli.StringFlag{Name: "circuit", Usage: "Transfer circuit (\"transfer\" / \"transfer-eddsa\")", Required: true},
```
sergeytimoshin: transfer, transfer-p256

prover/server/server/server.go
```diff
@@ -1108,29 +1090,15 @@ func (handler proveHandler) isBatchOperation(circuitType common.CircuitType) boo
...
 	switch circuitType {
-	case common.BatchUpdateCircuitType:
-		return "zk_update_queue"
-	case common.BatchAppendCircuitType:
-		return "zk_append_queue"
 	case common.BatchAddressAppendCircuitType:
 		return "zk_address_append_queue"
 	default:
-		return "zk_update_queue"
+		return ""
```
sergeytimoshin, line 1096: prob makes sense to add new circuits, otherwise they can be stucked in limbo in queues

PR #59, opened 2026-06-19
tests(photon): add more asserts to e2e test, add unit tests

sergeytimoshin:


ananas-block (dismissed):

sdk-libs/client/src/indexer.rs
```diff
@@ -405,4 +415,406 @@ mod tests {
...
+        assert_eq!(got.proofs[0].merkle_context.tree_type, 2);
+        assert_eq!(got.proofs[0].merkle_context.tree, tree);
+        assert_eq!(got.proofs[0].path, path);
+        assert_eq!(got.proofs[0].low_element, low);
+        assert_eq!(got.proofs[0].low_element_index, 3);
+        assert_eq!(got.proofs[0].high_element, high);
+        assert_eq!(got.proofs[0].high_element_index, 4);
+        assert_eq!(got.proofs[0].root, root);
+        assert_eq!(got.proofs[0].root_seq, 12);
+        assert_eq!(got.proofs[0].root_index, 13);
```
ananas-block: can we use a single assert_eq of the expected struct so that we ensure that the response is completely asserted?

ananas-block (approved):

PR #64, opened 2026-06-20
feat: cli use user_registry

swenhelius:
- switches to using user_registry for lookups, removes local
- adds cli config 
- loads user_registry program
- makes deposit to self the default
- allow transfer to non-user via internal lookup and fallback to withdraw
- more cleanups

### DIFF
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -378,7 +378,7 @@ pub(crate) struct TransferOptions {
 
     #[arg(
         long = "to",
-        help = "Recipient Solana pubkey looked up in the on-chain user registry",
+        help = "Recipient Solana pubkey; registered recipients receive a shielded transfer, unregistered recipients receive a public SOL withdrawal",
         value_name = "PUBKEY"
     )]
     pub(crate) to: String,
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -23,6 +23,12 @@ pub(crate) enum CliCommand {
     #[command(name = "start-prover", about = "Start the local prover server")]
     StartProver(StartProverOptions),
 
+    #[command(name = "config", about = "Show or update CLI configuration")]
+    Config {
+        #[command(subcommand)]
+        command: ConfigCommand,
+    },
+
     #[command(name = "wallet", about = "Private wallet commands")]
     Wallet {
         #[command(subcommand)]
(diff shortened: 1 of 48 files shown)


cli/src/args.rs
ananas-block, line 91: does balance sync the wallet as well?

swenhelius: yes

sdk-libs/client/src/actions/transaction.rs
```diff
@@ -14,15 +16,31 @@ pub struct ResolvedAddress {
...
-    fn resolve_address(&self, owner: Pubkey) -> Result<ResolvedAddress, ClientError>;
-}
-
 #[derive(Clone)]
 pub struct CreatedTransfer {
     pub signed: SignedTransaction,
     pub wait_tag: ViewTag,
-    pub recipient: ResolvedAddress,
+    pub recipient: TransferRecipient,
+    pub withdrawal: Option<TransactWithdrawal>,
```
ananas-block: shouldnt `withdrawal: Option<TransactWithdrawal>` be inside of `TransferRecipient::PublicWithdrawal` since it will always be required in that case ?

program-tests/shielded-pool/tests/localnet_wallet_cli_e2e.rs
```diff
@@ -133,7 +143,7 @@ fn wallet_cli_sol_cycle() -> Result<()> {
         "--airdrop-lamports",
         "20000000000",
     ])?;
-    let tree = parse_tree_pubkey(&create_tree_out)?;
+    let _tree = parse_tree_pubkey(&create_tree_out)?;
```
ananas-block: pls remove

program-tests/shielded-pool/tests/localnet_wallet_cli_e2e.rs
```diff
@@ -115,9 +124,10 @@ fn wallet_cli_sol_cycle() -> Result<()> {
     let alice = root.join("alice.pid.json");
     let bob = root.join("bob.pid.json");
     let tree_keypair = root.join("tree.json");
+    unsafe { env::set_var("ZOLANA_CONFIG", root.join("config.json")) };
```
ananas-block: why unsafe?

program-libs/interface/src/user_registry/state.rs
```diff
@@ -43,16 +43,23 @@ impl UserRecord {
...
-            _ => Err(borsh::io::Error::new(
-                borsh::io::ErrorKind::InvalidData,
-                "missing user record discriminator",
-            )),
+            _ => Err(invalid_user_record("missing user record discriminator")),
         }
     }
 
+    #[cfg(feature = "solana")]
+    pub fn try_from_account_checked(account: &solana_account::Account) -> borsh::io::Result<Self> {
```
ananas-block: this doesnt belong into the interface crate since it will never be used in a program, pls move to the client and remove the `solana-account` dependency.

program-libs/user-registry-interface/src/state.rs
ananas-block, line 1: we should have a separate `user-registry-interface` crate so that the shielded pool program doesnt pull in dependencies it doesn't need.

program-tests/shielded-pool/tests/localnet_wallet_cli_e2e.rs
swenhelius: actually no reason to . removed

program-libs/interface/src/user_registry/state.rs
swenhelius: done

program-libs/user-registry-interface/src/state.rs
swenhelius: ok, done

PR #65, opened 2026-06-20
feat: SPL support in CLI

swenhelius:
merge after #64 

- cli local asset registry
- test-mint command that create random mint, mints tokens to owner, adds asset to registry
- test cov for SPL flows in deposit transfer withdrawal

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4597,6 +4597,7 @@ dependencies = [
  "bytemuck",
  "cucumber",
  "groth16-solana",
+ "hex",
  "light-hasher",
  "num-bigint 0.4.6",
  "serde",
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -43,6 +43,12 @@ pub(crate) enum ConfigCommand {
 
     #[command(name = "set", about = "Update the CLI configuration file")]
     Set(ConfigSetOptions),
+
+    #[command(name = "asset-registry", about = "Show locally configured assets")]
+    AssetRegistry,
+
+    #[command(name = "add-asset", about = "Add or update a local SPL asset mapping")]
+    AddAsset(ConfigAddAssetOptions),
 }
 
 #[derive(Args, Debug, Clone)]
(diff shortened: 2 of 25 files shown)


ananas-block (commented):
couple comments

ananas-block:
Privacy: `transfer --to` downgrades shielded → public when the recipient isn't registered

In `create_transfer` (`sdk-libs/client/src/actions/transaction.rs`), if the recipient pubkey has no user-registry record, we fall back to a **public withdrawal** instead of a shielded transfer. So `wallet transfer --to <PUBKEY>` becomes a public, deanonymized withdrawal whenever the recipient hasn't registered yet. The only signal today is `mode=withdraw` in stdout.

Should we display a notice to the user before taking the public path?

ananas-block:
SPL transfer to an unregistered recipient withdraws to the sender's own token account

On the public-withdrawal fallback in `create_transfer` (`sdk-libs/client/src/actions/transaction.rs`), the SPL `user_token_account` is resolved from the **sender's** config (`configured_spl_token_account`). In `settle_spl` (`programs/shielded-pool/src/instructions/settlement/spl.rs`), the withdrawal CPI transfers `vault -> user_token_account`, and the `recipient` field is unused on that path. So the sender withdraws their shielded SPL back into their own token account while the output reads `to=<recipient>`. The recipient receives nothing.

The destination has to be a token account the recipient owns. This codebase has no ATA derivation (`test-mint` creates plain keypair token accounts), so the sender can't compute the recipient's token account. Options: reject the public fallback for SPL (require the recipient be registered), or take an explicit recipient token-account argument.

ananas-block:
CLI config file is re-read several times per command

`run_transfer` (`cli/src/wallet_cli/transaction.rs`) loads `CliConfigFile` from disk 4-5 times: once directly, again in `get_network`, and again in `sync_context` (each of which also calls `resolve_sync`, another load). Deposit and withdraw do the same. Suggest loading it once at the command entry and threading it through.

ananas-block (approved):

PR #72, opened 2026-06-22
refactor: spec to confidentiality

ananas-block:


tilo-14 (approved):
lgtm . we still need to update to rings

swenhelius (approved):

PR #66, opened 2026-06-20
chore: refactor key mgmt

swenhelius:
merge AFTER #70

PR #74, opened 2026-06-22
feat: merge service in client and cli

swenhelius:
merge after #66


ananas-block:
**`stop`/`status` miss a foreground `merge-service start`**

`pid_is_merge_service` requires the process command line to contain both `merge-service` and `run-loop`. Only the background worker's argv contains `run-loop` (`spawn_background_merge_service`); a foreground `start` runs `run_local_merge_service_loop` in the same process, with argv `merge-service start`, yet still writes a pid file via `MergeServicePidGuard::acquire`. For a foreground service the identity check then returns `false`:

- `stop_merge_service` skips `TERM`/`KILL` and only deletes the pid file. The process keeps running.
- `merge_service_running` reports not-running and deletes the live pid file.
- `MergeServicePidGuard::acquire` treats the live pid as stale and starts a second instance.

The README states `stop` handles a background or foreground process, so this is also a doc mismatch.

Fix: make the identity check match both spawn modes. Either gate on the `merge-service` subcommand marker alone, or route foreground `start` through the same `run-loop` entrypoint the check expects.

PR #81, opened 2026-06-25
refactor: make prover auto-download configurable

sergeytimoshin:

This PR is narrow: it only makes prover key auto-download configurable. It does not change the wallet transaction surface.

- `zolana start-prover --auto-download <bool>` controls whether the child prover may download missing proving keys.
- `zolana test-validator --prover-auto-download <bool>` forwards the same setting when the local validator launcher starts the prover; it has no effect with `--skip-prover`.
- `ZOLANA_PROVER_AUTO_DOWNLOAD` provides the same setting for both commands.
- The default remains `true`, preserving the current local-dev behavior.
- The prover still uses `ZOLANA_PROVER_KEYS_DIR` when set, otherwise `~/.config/zolana/proving-keys`.

## Why this exists

The CLI currently starts the prover in a convenience-first mode where missing keys can be fetched implicitly. That is good for local development, but awkward for CI, offline runs, preloaded-key environments, or production-shaped smoke tests where startup should fail fast instead of reaching the network. This keeps the convenient default while giving callers a deterministic `--auto-download false` / `ZOLANA_PROVER_AUTO_DOWNLOAD=false` path.

## Rebase notes

- Rebased onto current `main` (`2e1d7c81`).
- Resolved the only conflict in `cli/src/prover.rs` by preserving current main's offset `--metrics-address` behavior and threading this PR's configurable `--auto-download` through the same helper.
- Kept the diff scoped to `cli/src/args.rs`, `cli/src/localnet.rs`, and `cli/src/prover.rs`.

## Validation

- `cargo fmt -p zolana-cli`
- `cargo test -p zolana-cli -- --nocapture` (27 passed)

### DIFF
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -189,6 +189,16 @@ pub(crate) struct TestValidatorOptions {
     pub(crate) prover_port: u16,
 
     #[arg(
+        long = "prover-auto-download",
+        env = "ZOLANA_PROVER_AUTO_DOWNLOAD",
+        default_value_t = true,
+        action = ArgAction::Set,
+        value_parser = clap::builder::FalseyValueParser::new(),
+        help = "Allow the prover to download missing proving keys"
+    )]
+    pub(crate) prover_auto_download: bool,
+
+    #[arg(
         long,
         default_value_t = DEFAULT_PHOTON_PORT,
         help = "Photon indexer API port"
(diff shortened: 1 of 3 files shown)


ananas-block:
why do we need this?

sergeytimoshin:
Rebased this onto current `main` and resolved the conflict in `cli/src/prover.rs`.

Why we need it: the default local-dev path should keep auto-downloading missing proving keys, but CI/offline/preloaded-key runs need a deterministic fail-fast mode instead of silently reaching the network. This adds that opt-out through both launch paths:

- `zolana start-prover --auto-download false`
- `zolana test-validator --prover-auto-download false`
- `ZOLANA_PROVER_AUTO_DOWNLOAD=false`

I also kept current main's offset metrics-port behavior in the rebased prover argv helper and added focused tests for both the parser flags and generated prover args.

Validated locally with:

- `cargo fmt -p zolana-cli`
- `cargo test -p zolana-cli -- --nocapture` (27 passed)

ananas-block (approved):

PR #82, opened 2026-06-25
Add 40/250 batch address append key support

sergeytimoshin:
## Summary

- Add the regenerated 40/250 batch address append verifying key and export it from the batched Merkle tree verifier.
- Teach the prover key manager to load both 40/10 and 40/250 batch address append proving keys from the release bundle.
- Update proving-key release publishing/downloading to require the full key set, include the 40/250 asset, and handle split release assets for large keys.
- Add an xtask JSON verifying-key export path for regenerating vanilla Groth16 verifying-key Rust sources.
- Restore merge proof packing to the current main-branch 192-byte layout after the rebase.

## Validation

- cargo fmt
- cargo check --workspace
- cargo check -p spp-test-validator --test lifecycle
- cargo check -p xtask -p zolana-batched-merkle-tree
- bash -n prover/server/scripts/publish_keys_release.sh

## CI

- cargo check, clippy, and all SPP localnet jobs are passing on the latest run.
- tests / localnet photon / e2e was still running at the last check.


program-libs/batched-merkle-tree/src/verify/verifying_keys/batch_address_append_40_10.rs
```diff
@@ -1,65 +1,71 @@
 // This file is generated by xtask. Do not edit it manually.
 
 use groth16_solana::groth16::Groth16Verifyingkey;
+
 pub const VERIFYINGKEY: Groth16Verifyingkey = Groth16Verifyingkey {
-    nr_pubinputs: 1usize,
```
ananas-block: shouldnt we also generate a 250 inputs vk?

ananas-block:
most or all of the changes this pr introduces were merged as part of pr 89

PR #84, opened 2026-06-25
clean up spec api and deposit naming

sergeytimoshin:
## Summary

- Rename the spec's Photon-facing RPC method names to Helius-style lowerCamelCase while keeping Rust-shaped DTO examples in snake_case.
- Align the spec's proofless deposit instruction naming with the current `deposit` / `zone_deposit` code paths.
- Update transfer/user-flow wording from shield/unshield to deposit/withdraw where it is user-facing, while preserving shielded transfer/address terminology.

## Validation

- Searched `docs/spec.md` for stale snake_case RPC method names.
- Searched for legacy `proofless_shield` / `zone_proofless_shield` / unshield wording.
- Searched for accidental camelCase field names inside Rust-shaped DTO snippets.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -48,8 +48,8 @@
     - [Zone Accounts](#zone-accounts)
   - [Instructions](#instructions)
     - [transact](#transact)
-    - [proofless_shield](#proofless_shield)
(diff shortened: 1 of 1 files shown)


ananas-block:
we should probably reapply this

PR #93, opened 2026-06-30
feat: optional memo on output utxo serialization

ananas-block:
## Summary

Adds an optional **memo** to output UTXO serialization in the transaction crate.

The memo is modeled as a new `DataRecord::Memo(Vec<u8>)` variant in the shared `Data` struct (tag `3`, `u16` length prefix per the repo's wincode convention for byte vectors). Since every output plaintext — `TransferRecipientPlaintext`, `TransferSenderPlaintext`, and both `Anonymous*` variants — embeds `Data`, the memo flows into all of them automatically.

### Design: optional, uncommitted data

The on-chain commitment binds `data_hash` (derived from `utxo_data()`) and `zone_data_hash` (from `zone_data()`) only — **not** the full serialized `Data` blob. So a memo is encrypted into the recipient's note but is **not authenticated by the proof/commitment**: it is optional, informational data. This keeps the change contained entirely to the transaction crate, with no circuit or program changes.

## Changes

- **`data.rs`** — `DataRecord::Memo` variant; `validate()` extended to enforce canonical order `ZoneData < UtxoData < Memo` and reject duplicate memos; `memo()` accessor mirroring `zone_data()`/`utxo_data()`.
- **`instructions/transact/types.rs`** — `OutputUtxo::with_memo()` builder (sets no `data_hash`, reflecting it is uncommitted). Replaced the zone/non-zone toggle in `set_data_record` with a `canonical_data_order` helper — the old two-way `retain` would drop utxo data when a memo was added.
- **`serialization/anonymous.rs`** — Relaxed `AnonymousTransferRecipientPlaintext::into_utxo` to permit a memo-only `Data`, while still rejecting zone/utxo data with `UnsupportedOutputData`.

## Tests

- New unit tests: memo round-trip + readability, memo-only validity, duplicate rejection, non-canonical ordering, anonymous memo accept, anonymous zone/utxo reject.
- Extended the existing recipient round-trip cucumber step with memo-bearing `Data` variants.
- `cargo test -p zolana-transaction` green (42 scenarios + integration tests); `cargo clippy -p zolana-transaction --all-targets` clean.

### DIFF
--- a/cli/src/wallet_cli/deposit.rs
+++ b/cli/src/wallet_cli/deposit.rs
@@ -38,6 +38,7 @@ pub(super) fn run_deposit(opts: DepositOptions) -> Result<()> {
         asset,
         amount: opts.amount,
         spl_token_account,
+        memo: None,
     })?;
     let signature = deposit.send(&rpc, &material.funding, tree, &material.funding)?;
     wait_for_indexed_utxo(&indexer, deposit.view_tag(), signature)?;
--- a/program-libs/event/src/proofless.rs
+++ b/program-libs/event/src/proofless.rs
@@ -41,28 +41,3 @@ pub fn encode_verifiably_encrypted(blob: Vec<u8>) -> Vec<u8> {
     borsh::to_vec(&OutputData::VerifiablyEncrypted(blob))
         .expect("shielded-pool output data serialization is infallible")
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn proofless_output_memo_round_trips() {
-        let output = ProoflessOutput {
-            owner: [1u8; 32],
-            blinding: [2u8; 31],
-            asset: [3u8; 32],
-            amount: 99,
-            data_hash: None,
-            utxo_data: None,
-            zone_program_id: None,
-            zone_data_hash: None,
-            zone_data: None,
-            memo: Some(b"hello".to_vec()),
-        };
-        let bytes = borsh::to_vec(&output).unwrap();
-        let parsed = ProoflessOutput::try_from_slice(&bytes).unwrap();
-        assert_eq!(parsed, output);
-        assert_eq!(parsed.memo.as_deref(), Some(b"hello".as_slice()));
-    }
-}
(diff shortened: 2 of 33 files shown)


program-libs/event/src/proofless.rs
```diff
@@ -38,3 +41,28 @@ pub fn encode_verifiably_encrypted(blob: Vec<u8>) -> Vec<u8> {
...
+            zone_data_hash: None,
+            zone_data: None,
+            memo: Some(b"hello".to_vec()),
+        };
+        let bytes = borsh::to_vec(&output).unwrap();
+        let parsed = ProoflessOutput::try_from_slice(&bytes).unwrap();
+        assert_eq!(parsed, output);
+        assert_eq!(parsed.memo.as_deref(), Some(b"hello".as_slice()));
+    }
+}
```
sergeytimoshin: is it necessary?

cli/src/wallet_cli/deposit.rs
```diff
@@ -38,6 +38,7 @@ pub(super) fn run_deposit(opts: DepositOptions) -> Result<()> {
         asset,
         amount: opts.amount,
         spl_token_account,
+        memo: None,
```
sergeytimoshin, line 41: we may pass memo as cli arg on deposit

program-libs/interface/src/instruction/instruction_data/deposit.rs
```diff
@@ -79,3 +87,24 @@ impl ZoneDepositIxData {
...
+            public_amount: Some(1_000),
+            utxo_data: None,
+            memo: Some(b"hello memo".to_vec()),
+        };
+        let bytes = data.serialize().unwrap();
+        let parsed = DepositIxData::deserialize(&bytes).unwrap();
+        assert_eq!(parsed, data);
+        assert_eq!(parsed.memo.as_deref(), Some(b"hello memo".as_slice()));
+    }
+}
```
sergeytimoshin: is it necessary?

sergeytimoshin (approved):

PR #97, opened 2026-07-02
feat(sdk): client DevX — idempotent ATA, one-call submit, Wallet-owned registry, ensure_registered, Transaction::from_wallet

tilo-14:
## Summary

A batch of `zolana-client` / `zolana-transaction` developer-experience improvements so app and example code stops hand-rolling instruction plumbing. No program or circuit changes — SDK, CLI, and test call-sites only.

## What's in it

- **Idempotent ATA builder + action** — `zolana_interface::instruction::CreateAssociatedTokenAccount` (emits the SPL `CreateIdempotent` layout) and `zolana_client::create_associated_token_account(rpc, payer, owner, mint) -> (Signature, Pubkey)`. Replaces hand-rolled ATA instructions.

- **One-call private-transaction submit** — `Rpc::get_spend_proofs` (production polling proof-fetch, the non-panicking replacement for the test-utils `wait_for_*` retry) + `Submit { signed, withdrawal, cu_limit }.execute(rpc, prover, payer, tree)`. Collapses the ~45-line fetch→assemble→prove→build→send sequence. `.submit()` sugar on `CreatedTransfer` / `CreatedWithdrawal` mirrors `Deposit::send`.

- **`AssetRegistry` moves onto `Wallet`** — dropped from the client-facing signatures (`sync_wallet`, `get_private_token_balances`, `CreateTransfer`/`CreateWithdrawal`, `sign_transaction`). `Wallet::new(keypair, registry)`; the pure builder/serialization tier still takes `&AssetRegistry` (it's a genuine input there).

- **`ensure_registered`** — idempotent user-registry publish (register / update_keys / no-op) in `client::user_registry`, the write counterpart to the existing read helpers. CLI `register_wallet_on_chain` now delegates to it, removing the duplicated derive machinery.

- **`Transaction::from_wallet(&wallet, inputs, payer)`** — builds a client transaction from a wallet's UTXOs, wrapping the existing `Transaction::new` free constructor (the established local idiom).

## API-surface changes (breaking)

| Before | After |
|---|---|
| `Wallet::new(keypair)` | `Wallet::new(keypair, registry)` |
| `sync_wallet(wallet, indexer, assets)` | `sync_wallet(wallet, indexer)` |
| `get_private_token_balances(wallet, assets)` | `get_private_token_balances(wallet)` |
| `CreateTransfer { .., assets }` / `CreateWithdrawal { .., assets }` | field removed |
| `sign_transaction(tx, owner, authority, assets)` | `sign_transaction(tx, wallet, owner, authority)` |

## Testing

- `cargo build --workspace --all-targets`, `cargo clippy --workspace --all-targets`, `cargo +nightly fmt --check` — clean.
- `just test-sdk-libs` — pass, including new unit tests for `get_spend_proofs`, `Submit::prepare`, `ensure_registered` (register/update/noop), `Transaction::from_wallet`, and the ATA builder.
- Full prove→submit and localnet e2e is exercised by the example crate, which now lives in the separate `zolana-examples` repo (depending on this SDK by path).

## Notes

- The examples were added on this branch then moved out to `zolana-examples`; the net diff contains no examples.
- One pre-existing prover-key failure (P256 `constraint #65096`) reproduces on clean `main` and is unrelated to this change.


sdk-libs/transaction/src/wallet/state.rs
```diff
@@ -110,10 +114,14 @@ pub struct Wallet {
 }
 
 impl Wallet {
-    pub fn new(keypair: ShieldedKeypair) -> Result<Self, TransactionError> {
+    pub fn new(
+        keypair: ShieldedKeypair,
+        registry: AssetRegistry,
```
ananas-block, line 124: what if the registry needs to be kept in sync with newly registered assets?

sergeytimoshin: btw, we might add asset registry to the indexer to be able to fetch all pool assets and or assets by id

tilo-14: it'd make sense to move to indexer, because otherwise we have the extra field in the api surface and this way we also ensure sync f the asset registry

swenhelius:
can we remove these two features and do them in a separate PR:

- One-call private-transaction submit 
- Transaction::from_wallet(&wallet, inputs, payer)

prod users won't be able to use these two as is, but the rest LGTM @tilo-14

swenhelius (approved):

PR #99, opened 2026-07-03
feat: forester info command

sergeytimoshin:

Adds the read-only forester status command:

```bash
forester info [--tree <POOL_TREE>] [--json]
```

- Reads `RPC_URL` from the environment and inspects the selected pool tree, defaulting to the devnet/localnet SPP tree fixture.
- Optionally reads `PAYER` to show the forester fee-payer balance and `PROVER_URL` to show which prover is configured.
- Reports state-tree height/leaves/root, nullifier-queue batch state, queued counts, ready-to-forest zkp-batches, inserted zkp-batches, queue pointers, and nullifier root.
- Emits a human-readable report by default and a stable JSON object with `--json` for monitoring or for gating `forester run`.
- Never proves or submits transactions; this PR is the safe inspection layer that `#101` builds on.

## Recent update in this PR

- Fixes the stale `batch_update` account-count test expectation from 3 to 4.
- Keeps the command intentionally narrow: it answers "is there work for the forester, and is the forester funded/configured?" without mutating chain state.

## Validation / status

- Forester compile check is green on the latest run shown for `#99`.
- Rust CI for `#99` is green on the latest run shown, including `tests / cli` and the localnet photon jobs.


forester/Dockerfile
```diff
@@ -1,20 +1,37 @@
...
     pkg-config \
     libssl-dev \
+    git \
+    ca-certificates \
+    protobuf-compiler \
     && rm -rf /var/lib/apt/lists/*
 
+# Whole workspace as context (kept small by zolana-main/.dockerignore, which
+# excludes target/, .git/, proving-keys/, ...).
 COPY . .
```
ananas-block, line 29: These comments point at infrastructure that isn't in this repo. There's no `docker-bake.hcl` (so no `forester` target, no `context = ./zolana-main`), no `zolana-main/` directory, and `prover/server/docker-compose.yml` has no `x-photon-image`.

The `COPY . .` also assumes the workspace root as build context, but there's no root `.dockerignore`. `forester/.dockerignore` isn't honored when the context root is the workspace root, so `target/` gets pulled into the build context along with everything else it lists.

Add a root `.dockerignore` (or `forester/Dockerfile.dockerignore` for BuildKit) so the exclusions apply, and fix the comment to describe the real in-repo setup. If the bake/`zolana-main` orchestration lives in a separate ops repo, say that instead of citing paths that resolve to nothing here.

forester/src/info.rs
```diff
@@ -0,0 +1,227 @@
...
+
+fn lamports_to_sol(lamports: u64) -> String {
+    format!(
+        "{}.{:09}",
+        lamports / LAMPORTS_PER_SOL,
+        lamports % LAMPORTS_PER_SOL
+    )
+}
+
+fn hex(bytes: &[u8; 32]) -> String {
```
ananas-block: `hex` is already a workspace dependency, and `hex::encode` is used across `cli`, `sdk-libs`, and `program-tests`. This version duplicates it and allocates a `String` per byte via `push_str(&format!(...))`. Drop the helper and call `hex::encode(&state_root)` / `hex::encode(&root)`.

forester/src/cli.rs
```diff
@@ -1,4 +1,9 @@
 use clap::{Parser, Subcommand};
+use solana_pubkey::Pubkey;
+
+/// Default SPP pool tree the forester maintains (devnet/localnet fixture).
+pub const DEFAULT_TREE: Pubkey =
```
ananas-block, line 5: This is the 4th hardcoded copy of `treeYbr45LjxovKvtD46uEphM64kwoFFPYhVNw1A8x8` -- it's also in `cli/src/cli_config.rs`, `cli/src/wallet_cli/resolve.rs`, and `xtask/src/init_protocol.rs`. Hoist the fixture address into `zolana-interface` and reference it from all four callers so it can't drift.

forester/src/info.rs
```diff
@@ -0,0 +1,227 @@
...
+        let metadata = *nullifier_tree.get_metadata();
+        let batches = &metadata.queue_batches;
+        let mut infos = Vec::with_capacity(batches.batches.len());
+        let mut ready_total = 0u64;
+        for (index, batch) in batches.batches.iter().enumerate() {
+            let ready = batch.get_num_ready_zkp_updates();
+            ready_total += ready;
+            infos.push(BatchInfo {
+                index,
+                state: state_name(batch.get_state()),
```
ananas-block: `Batch::get_state()` does `self.state.into()`, and `From<u64> for BatchState` calls `panic!("Invalid BatchState value")` for anything other than 0/1/2. `state` comes from an account whose pubkey the user supplies via `--tree`. `TreeAccount::from_bytes` validates the UTXO layout first, so a wrong account usually errors before batch parsing -- but a corrupt account that matches the layout would abort the process instead of returning an error. Map an out-of-range state to an `"unknown"` label or a returned error rather than letting the shape check be the only thing standing between user input and a panic.

PR #101, opened 2026-07-03
feat: forester run — prove & submit ready nullifier-tree zkp-batch updates via the smart-account vault

sergeytimoshin:

Adds the mutating forester worker command, stacked on `#99`:

```bash
forester run [--tree <POOL_TREE>] [--settings <SMART_ACCOUNT_SETTINGS>] [--account-index <N>] [--dry-run] [--max-batches <N>] [--watch] [--poll-secs <N>]
```

- Reads `RPC_URL`, `PHOTON_URL`, `PROVER_URL`, and `PAYER` from the environment for submit mode.
- `--dry-run` only performs the read/RPC/Photon/reconstruction/root-match path and does not require `PROVER_URL`, `PAYER`, or `--settings`.
- Submit mode requires `--settings`; the vault at `--account-index` is the tree's `forester_authority`, while `PAYER` signs as a member of that smart account.
- Reconstructs the reference indexed nullifier tree from Photon's ordered `get_nullifier_queue_elements` response, then refuses to proceed unless the reconstructed root equals the on-chain nullifier root.
- Proves ready zkp-batches on the forester prover and submits `batch_update_nullifier_tree` through the smart-account vault via `execute_sync`.
- Submits zkp-batches sequentially because each update checks `old_root == current_root`; `--max-batches` caps a run, and `--watch` keeps polling for newly-ready batches.

## Recent updates in this PR

- Switches the Photon path to the generated `zolana-api` `BlockingZolanaApi` client instead of the earlier bespoke RPC client.
- Promotes the smart-account PDA / `execute_sync_ix` helper from `zolana-test-utils` into a reusable `zolana-smart-account-client` SDK crate; test-utils now re-export it.
- Keeps key rotation out of scope; this branch is based on current main's batch/nullifier key state rather than carrying a separate proving-key rotation.

## Stacking / dependencies

- Stacked on `#99` (`forester info`).
- Depends on the Photon API support for `get_nullifier_queue_elements`; intended merge order is Photon privacy API support -> `#102` zolana-api regeneration -> this PR.

## Validation / status

- Unit tests pin the reconstructed empty nullifier tree root and root chaining across zkp-batches.
- Ignored localnet tests cover dry-run plumbing and a `FORESTER_BIN`-gated full drain with the real binary.
- Local validation noted in the original PR: the full drain forested 20 zkp-batches through the smart-account vault.
- Latest CI shown for `#101`: forester compile, rustfmt, cargo-machete, clippy, cargo check, CLI tests, SDK tests, program tests, and most localnet photon jobs are green; `tests / localnet photon / zone lifecycle` is currently failing and still needs follow-up/triage.


sergeytimoshin:
Now depends on #102 (regenerate zolana-api + `get_nullifier_queue_elements`): the second commit switches the forester off its bespoke photon client to `zolana-api`'s `BlockingZolanaApi`. Merge order: photon-privacy#15 → #102 → #101. (CI here is red until #102 lands.)

forester/src/run.rs
```diff
@@ -0,0 +1,528 @@
...
+/// configured forester authority.
+fn forester_keypair() -> Result<Keypair> {
+    let payer = env::var("PAYER").context("PAYER is not set (forester signing keypair)")?;
+    let bytes: Vec<u8> =
+        serde_json::from_str(&payer).context("PAYER must be a JSON byte array keypair")?;
+    Keypair::try_from(bytes.as_slice())
+        .map_err(|err| anyhow!("PAYER is not a valid keypair: {err}"))
+}
+
+fn hex(bytes: &[u8; 32]) -> String {
```
ananas-block: nit: `hex()` is duplicated byte-for-byte in info.rs. Leaving it as a local helper is fine; add a cross-reference so the two don't drift.

forester/src/run.rs
```diff
@@ -0,0 +1,528 @@
...
+fn path_to_biguint(path: Vec<[u8; 32]>) -> Vec<BigUint> {
+    path.into_iter()
+        .map(|node| BigUint::from_bytes_be(&node))
+        .collect()
+}
+
+/// Resolve the forester signing keypair from `PAYER` (a JSON byte array, as in
+/// `info`). Required for `run`: the update must be signed by the tree's
+/// configured forester authority.
+fn forester_keypair() -> Result<Keypair> {
```
ananas-block, line 470: nit: the `PAYER` byte-array parse is duplicated: `run::forester_keypair` requires it, `info::read_forester` tolerates it being unset. Keep them in sync if the keypair encoding changes.

forester/src/run.rs
```diff
@@ -0,0 +1,528 @@
...
+        let signature = batch_update_nullifier_tree_once(ForestParams {
+            rpc_url,
+            member,
+            settings,
+            account_index,
+            pool_tree: tree,
+            batch_update: BatchUpdateNullifierTreeData {
+                new_root,
+                old_root,
+                zkp_batch_index: zkp_index as u16,
```
ananas-block, line 324: nit: `zkp_index` is u64; this narrows to u16 silently. It fits, since it's bounded by the batch's num_zkp_batches (batch_size / zkp_batch_size). A comment stating that bound would document why the cast can't truncate.

sdk-libs/zolana-api/src/lib.rs
```diff
@@ -154,7 +156,7 @@ impl ZolanaApi {
             method: types::PostGetEncryptedUtxosByTagsBodyMethod::GetEncryptedUtxosByTags,
             params: types::PostGetEncryptedUtxosByTagsBodyParams {
                 cursor,
-                limit: limit.map(types::Limit),
+                limit: limit.and_then(|value| ::std::num::NonZeroU64::new(value).map(types::Limit)),
```
ananas-block, line 159: behavior note (also at 188/363/392): `Limit` now wraps `NonZeroU64`, so `Some(0)` is dropped and sent as no limit rather than a zero-row request. The regenerated type forces this and a 0 limit is meaningless, so it stands. It changes behavior versus the old `limit.map(types::Limit)`.

ananas-block (commented):
A few nits and one behavior note on the forester run flow. Nothing blocking. The reconstruct-then-root-match safety net, sequence-gap checks, and the fresh-snapshot-per-drain recovery model are solid, and the e2e coverage (dry-run smoke + FORESTER_BIN full drain) is good.

One more (couldn't inline it, the line is unchanged context): `main.rs` `Commands::Start` is a no-op stub. `run --watch` covers the drain loop, so this looks like a deliberate placeholder for the always-on worker. Confirm that's intended, and consider a comment saying the worker is future work.

swenhelius (approved):

PR #104, opened 2026-07-03
feat: permissionless spl interface creation

ananas-block:
## Summary

- Add `spl_interface_creation_is_permissionless` to `ProtocolConfig` (131 -> 132 bytes). When set, any signer can call `create_spl_interface`; otherwise it remains gated by `protocol_authority`.
- Add `UpdateProtocolConfigData::SplInterfaceCreationPermissionless` (appended, existing borsh variant indices unchanged) and wire it through `create_protocol_config`, the builders, and the program-test admin helpers.
- Add an `update-protocol-config` xtask: reads and prints the on-chain config, resolves the protocol Squads settings by scanning smart-account seeds for the vault matching `protocol_authority`, and sends flag updates wrapped in `execute_sync`.
- Update `docs/spec.md` (ProtocolConfig struct, tag 4 gating).

## Devnet state

The devnet config account was already migrated to the new 132-byte layout with a temporary resize path in `update_protocol_config` (sig `2NeHFo4Tin6DwcuZRNewdnMTJcid2GYycdV33EurwVZyotiDoTQDxkCrRMeBgyFKfaKm3MKYdWd8rQCj6wAmDpUM`), and `spl_interface_creation_is_permissionless` is now `true` there. The resize code is removed again in this branch; the currently deployed devnet program still contains it and should be redeployed from this branch.

## Testing

- New BDD scenario: a non-authority registers an SPL interface once the flag is enabled.
- Full shielded-pool BDD suite passes (33 scenarios, 151 steps), plus workspace check, clippy, and fmt.

### DIFF
--- a/cli/src/wallet_cli/tree.rs
+++ b/cli/src/wallet_cli/tree.rs
@@ -42,6 +42,7 @@ pub(super) fn run_create_tree(opts: CreateTreeOptions) -> Result<()> {
             forester_authority: authority_address,
             zone_creation_authority: authority_address,
             zone_creation_is_permissionless: false,
+            spl_interface_creation_is_permissionless: false,
         }
         .instruction();
         let signature =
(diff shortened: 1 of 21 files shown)


tilo-14 (approved):
lgtm

PR #105, opened 2026-07-03
fix(prover): fetch proving keys from the GitHub release, drop GCS

sergeytimoshin:
## What
The prover key downloader still routed all keys except batch address-append through the light-protocol **GCS bucket** (`DefaultBaseURL` + `DownloadKey`). rings/zolana doesn't use GCS — Zolana keys live only on the private `helius-labs/zolana` GitHub release — so auto-downloading transfer/merge keys fetched from a bucket that never holds them.

## Changes
- **Route every key through the release path**; delete the GCS code (`DefaultBaseURL`, `DownloadKey`, `downloadChecksum`, `downloadFileWithResume`, `DownloadConfig.BaseURL`, the `--download-url` flag, `usesProvingKeysRelease`).
- **Download via the GitHub REST API + bearer token** (`GITHUB_TOKEN`/`GH_TOKEN`) instead of the `gh` CLI, so auto-download works in the **distroless** prover image (which has no `gh`).
- Release tag overridable via `PROVING_KEYS_RELEASE_TAG` (key rotations without a rebuild).
- Split assets (`.key.part-*`) still assembled + checksum-verified.

## Notes
- The proving-keys repo is **private**, so auto-download requires a token with read access; without one the API returns `404` with a hint. Deployments that **bake** keys into the image need no token and never hit this path.

## Testing
- `go build ./...` + `go vet` clean.
- Verified against the private `transfer-keys-v10` release: with a token it fetched + checksum-verified a key; without a token it 404s with the hint.

### DIFF
--- a/prover/server/prover/common/key_downloader.go
+++ b/prover/server/prover/common/key_downloader.go
@@ -3,37 +3,36 @@ package common
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 	"zolana/prover/logging"
 )
 
 const (
-	DefaultBaseURL       = "https://storage.googleapis.com/light-protocol-proving-keys/light-protocol-keys"
 	DefaultMaxRetries    = 10
 	DefaultRetryDelay    = 5 * time.Second
 	DefaultMaxRetryDelay = 5 * time.Minute
 
-	// ProvingKeysRepo and ProvingKeysReleaseTag identify the GitHub release
-	// that hosts Zolana-specific proving keys and their CHECKSUM. They must
-	// match the repo/tag used by scripts/publish_keys_release.sh. The repo is
-	// private, so the keys are fetched with the `gh` CLI (which carries the
-	// caller's auth, or CI's same-repo GITHUB_TOKEN) rather than an
-	// unauthenticated URL.
-	ProvingKeysRepo       = "helius-labs/zolana"
-	ProvingKeysReleaseTag = "transfer-keys-v10"
+	// ProvingKeysRepo and ProvingKeysReleaseTag identify the GitHub release that
+	// hosts every Zolana proving key and its CHECKSUM. They must match the
+	// repo/tag used by scripts/publish_keys_release.sh. The repo is private, so
+	// assets are fetched from the GitHub REST API with a bearer token from
+	// GITHUB_TOKEN / GH_TOKEN -- this works in the distroless prover image, which
+	// has no `gh` CLI. The tag is overridable via PROVING_KEYS_RELEASE_TAG so key
+	// rotations don't require a prover rebuild.
+	ProvingKeysRepo             = "helius-labs/zolana"
+	ProvingKeysReleaseTag       = "transfer-keys-v10"
+	provingKeysReleaseTagEnvVar = "PROVING_KEYS_RELEASE_TAG"
 )
 
 type DownloadConfig struct {
-	BaseURL       string
 	MaxRetries    int
 	RetryDelay    time.Duration
 	MaxRetryDelay time.Duration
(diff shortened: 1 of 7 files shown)


prover/server/prover/common/key_downloader.go
```diff
@@ -170,254 +108,150 @@ func calculateBackoff(attempt int, initialDelay, maxDelay time.Duration) time.Du
...
+// listReleaseAssets fetches the asset list for the configured release tag. The
+// proving-keys repo is private, so the request carries a bearer token.
+func listReleaseAssets() ([]releaseAsset, error) {
+	tag := releaseTag()
+	url := fmt.Sprintf("https://api.github.com/repos/%s/releases/tags/%s", ProvingKeysRepo, tag)
+	req, err := githubAPIRequest(http.MethodGet, url, "application/vnd.github+json")
+	if err != nil {
+		return nil, err
+	}
+	resp, err := http.DefaultClient.Do(req)
```
ananas-block: `listReleaseAssets` uses `http.DefaultClient`, which has no timeout, while the asset download uses a 60-minute client. A stalled metadata connection blocks with no bound. Give this call an explicit timeout.

prover/server/prover/common/key_downloader.go
```diff
@@ -170,254 +108,150 @@ func calculateBackoff(attempt int, initialDelay, maxDelay time.Duration) time.Du
...
-				continue
-			}
-			return fmt.Errorf("failed to download after %d attempts: %w", config.MaxRetries, err)
+			lastErr = err
+			continue
 		}
-
-		// Check response status
-		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+		if resp.StatusCode != http.StatusOK {
```
ananas-block, line 454: A permanent 401/403/404 on the asset GET retries 10 times with backoff, ~25 minutes before failing. Return immediately on 4xx, matching the auth-hint handling in `listReleaseAssets`.

prover/server/prover/common/key_downloader.go
```diff
@@ -447,8 +281,8 @@ func EnsureProvingKeyFromRelease(keyPath string, autoDownload bool) error {
 	logging.Logger().Info().
 		Str("file", filename).
 		Str("repo", ProvingKeysRepo).
-		Str("tag", ProvingKeysReleaseTag).
-		Msg("Downloading proving key from GitHub release via gh")
+		Str("tag", releaseTag()).
+		Msg("Downloading proving key from GitHub release via REST API")
```
ananas-block, line 547: Two issues in this function's download path:

1. **CHECKSUM re-downloaded, release listed twice per key.** The two `downloadReleaseAsset` calls below each fetch the asset list, and the same CHECKSUM is re-downloaded for every key. Preloading N keys makes ~2N list calls and N identical CHECKSUM downloads; the old `downloadChecksum` cache avoided this. Fetch the asset list once and reuse it.

2. **Release CHECKSUM overwrites the shared `proving-keys/CHECKSUM`.** `downloadReleaseAsset("CHECKSUM", dir)` writes the release manifest into the keys dir, dropping entries for locally-generated keys (e.g. squads keys hashed by `generate_checksums.py`). The `gh --clobber` path did the same, but merge/batch/merkle keys now route through here too, so it triggers on more startups. Confirm the prover uses a dedicated keys dir, or write the release CHECKSUM to a release-specific filename.

ananas-block (commented):
Review of the GitHub-release key downloader. The direction is right and the happy path looks correct; the redirect/token-stripping note on `downloadAssetByID` is accurate. `go build ./...` and `go vet` are clean. Three findings that don't sit on diff lines, plus inline comments below.

**`main.go:411` / `:505` — `--max-retries` and `--download-max-retries` are no-ops.** `downloadAssetByID` hardcodes `DefaultMaxRetries`/`DefaultRetryDelay`/`DefaultMaxRetryDelay`. Nothing reads `DownloadConfig.MaxRetries`, `RetryDelay`, or `MaxRetryDelay` anymore, yet both subcommands still expose the flags, assign them, and log the value. `download --max-retries 3` runs 10 retries. Either thread the config through `EnsureProvingKeyFromRelease` -> `downloadReleaseAsset` -> `downloadAssetByID`, or drop the three fields, both flags, and the log line.

**`key_downloader.go:370` — delete the `EnsureTransferKeyFromRelease` shim.** It only forwards to `EnsureProvingKeyFromRelease`, and the sole caller is `lazy_key_manager.go:235` (no test references it). Call `EnsureProvingKeyFromRelease` directly there and remove the function.

**`key_downloader.go:299` — split parts left on disk.** After `assembleReleaseAssetParts` renames the key into place, the `.part-*` files stay in the keys dir. Remove them once assembly succeeds.

PR #106, opened 2026-07-04
feat(prover): async queue for transfer/merge proofs

sergeytimoshin:
## What
Only batch **address-append** proofs were routed to the Redis async queue; **transfer/merge** circuits ran synchronously. Under concurrency (a soak/bench, or many wallet clients) each transfer spawns a ~1.5 GB Groth16 prove goroutine at once, so a single prover can OOM or stall.

## Change
- Route transfer/merge circuits to a dedicated **`zk_transfer_queue`** and add a **`TransferQueueWorker`** that drains it with bounded concurrency (`TRANSFER_WORKER_CONCURRENCY`, defaulting to the shared `getMaxConcurrency`).
- Make `GetQueueNameForCircuit` the source of truth for which circuits are queued; `shouldUseQueueForCircuit` derives from it.
- Reuse the same key manager and proof implementations as the synchronous handler.
- Start the transfer worker for the full server and for `--circuit transfer` queue workers.
- Forward Redis configuration through the Rust test launcher and fix the CLI's Go boolean flag encoding so `--redis-url` is not dropped.

## Why
This decouples client concurrency from prover capacity: clients submit freely while the worker drains at a sustainable rate. It enables high-throughput tests without stampeding and protects a shared prover in production.

## Client side
Clients poll jobs through the generic `ProverClient` job-envelope auto-detection from #107, the same mechanism used for address-append. No transfer-specific polling path is needed.

**Compatibility:** a Redis-enabled prover now returns a job handle for transfers, so clients must use the async-aware `ProverClient` from #107. A prover without Redis continues serving transfers synchronously.

## Testing
- Added `TestGetQueueNameForCircuit` for routing coverage.
- Added the required `async prover / transfer queue` check with a real Redis service.
- The integration check starts the CLI-managed prover, submits a real `2x3` transfer through `/prove`, polls `/prove/status` through the Rust client, verifies that `TransferQueueWorker` advanced `zk_results_queue`, and verifies the Groth16 proof against the committed key.
- Local verification: workspace clippy, client and CLI unit suites, Go routing test, workflow lint, and the real Redis-backed proof all pass.

### DIFF
--- a/.github/workflows/async-prover.yml
+++ b/.github/workflows/async-prover.yml
@@ -1,0 +1,55 @@
+name: async prover
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  transfer-queue:
+    name: async prover / transfer queue
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    timeout-minutes: 30
+    services:
+      redis:
+        image: redis:7.4.4-alpine3.21
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 2s
+          --health-timeout 5s
+          --health-retries 10
+    env:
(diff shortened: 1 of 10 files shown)


ananas-block:
do test this in ci?

swenhelius:
> do we test this in ci?

no, we still need to use the redis queue mode with TransferQueueWorker and using the async /prove endpoint for transfers as a CI test I think @sergeytimoshin

PR #107, opened 2026-07-04
feat(client): async-aware ProverClient (poll queued proofs)

sergeytimoshin:
## What
`zolana-client`'s `ProverClient` was synchronous: `send` POSTs `/prove` and parses the body as a proof. A Redis-enabled prover answers heavy circuits **asynchronously** — it returns a job handle (`{job_id, status, status_url}`, HTTP 202) instead of a proof — which the sync client can't parse.

## Change
`ProverClient::send` now auto-detects the async job envelope (a `job_id` with no `proof`) and polls `/prove/status?job_id=…` (3 s interval, 1200 s bound) until `status:"completed"`, extracting the proof from the `{proof, proof_duration_ms}` result. A synchronous prover still returns the proof directly, so the sync path is unchanged.

## Why this is its own PR
This is the shared client half of async proving. Both build on it:
- **#101** (forester `run`) needs it for batch address-append (that PR currently carries an identical inline copy — it can drop it once this lands).
- **#106** (prover async transfer/merge queue) needs it so the **CLI** (`cli/src/wallet_cli/transaction.rs` → `ProverClient::prove_transfer*`) transparently polls queued transfer proofs.

Generic by design — it keys off the response shape, not the circuit — so it covers address-append, transfer, and merge with no per-circuit client code.

## Testing
- `cargo check -p zolana-client` clean.
- Exercised end-to-end on devnet via the forester (batch address-append: submit → poll → proof → on-chain).

### DIFF
--- a/sdk-libs/client/src/prover/client.rs
+++ b/sdk-libs/client/src/prover/client.rs
@@ -63,6 +63,12 @@ const PROVE_RETRY_BACKOFF_SECS: u64 = 2;
 // well before this.
 const PROVE_REQUEST_TIMEOUT_SECS: u64 = 600;
 const PROVE_CONNECT_TIMEOUT_SECS: u64 = 10;
+// Redis-backed provers queue heavy batch proofs (address-append) and return a
+// job handle immediately instead of blocking; the client then polls the status
+// endpoint. The first batch job loads a multi-GB proving key before proving, so
+// the bound is generous — the status endpoint itself returns immediately.
+const ASYNC_POLL_INTERVAL_SECS: u64 = 3;
+const ASYNC_MAX_WAIT_SECS: u64 = 1200;
 
 fn build_http_client() -> reqwest::blocking::Client {
     reqwest::blocking::Client::builder()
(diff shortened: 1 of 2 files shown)


sdk-libs/client/src/prover/client.rs
```diff
@@ -63,6 +63,12 @@ const PROVE_RETRY_BACKOFF_SECS: u64 = 2;
 // well before this.
 const PROVE_REQUEST_TIMEOUT_SECS: u64 = 600;
 const PROVE_CONNECT_TIMEOUT_SECS: u64 = 10;
+// Redis-backed provers queue heavy batch proofs (address-append) and return a
+// job handle immediately instead of blocking; the client then polls the status
+// endpoint. The first batch job loads a multi-GB proving key before proving, so
+// the bound is generous — the status endpoint itself returns immediately.
+const ASYNC_POLL_INTERVAL_SECS: u64 = 3;
+const ASYNC_MAX_WAIT_SECS: u64 = 1200;
```
ananas-block: we should make this configurable same as we have in light-client

ananas-block (approved):

PR #112, opened 2026-07-06
feat: spec SLIP-0010 key derivation + create_private_wallet onboarding

tilo-14:
## Summary

Embedded-wallet onboarding for the SDK, in two parts. Stacked on #109 so the examples get one lineage (`main` ← #108 ← #109 ← this).

**Part 1 — `zolana-keypair`: spec SLIP-0010-P256 derivation.** Implements the spec's wallet key hierarchy (`docs/spec.md`, "Signing Key" / "ViewingKey"): `SigningKey::from_seed` (`m/44'/TSPP_COIN_TYPE'/account'/0'/0'`), `ViewingKey::from_seed` (`.../1'/0'`), and `ShieldedKeypair::from_seed` (both siblings + the nullifier key from the derived signing secret). One BIP-39 seed reproduces the wallet's shielded keys — mnemonic-recoverable. Verified against the official SLIP-0010 nist256p1 test vectors, including both invalid-key retry branches.

**Part 2 — `zolana-client`: `create_private_wallet(rpc, owner, keypair, registry)`.** The onboarding call every integrator writes, composed from the existing primitives: `ensure_registered` + `Wallet::new`. Replaces the example-crate helper of the same name, with two corrections:
- The caller supplies the `ShieldedKeypair` — the SDK never reads `owner.secret_bytes()` (same input-model rule that kept `Transaction::from_wallet` out in #108).
- `owner` is named for what it is: the registered on-chain identity (record PDA keyed under its address, must sign). It is not a third-party fee payer; nobody can register on someone else's behalf.

## Use case: embedded wallets

The app custodies a per-user BIP-39 `wallet_seed` and derives the full shielded keypair from it. Precisely: one seed reproduces the *shielded keypair*; the Solana identity (`owner`) is a separate secret (apps can derive both from one mnemonic — TSPP paths + Solana's `m/44'/501'`). Self-custody flows construct the `ShieldedKeypair` from their own key management and call the same function. `from_seed` wallets sign on the P256 rail; `from_ed25519` (Solana-keypair-only, caller-provided viewing key) stays on the eddsa rail — the two are different wallets, not encodings of one.

## Spec gap (flagged, no spec edit)

`docs/spec.md` declares `TSPP_COIN_TYPE = 1445561917' (placeholder)` but its own formula `SHA-256("luminous.TSPP.v1")[0..4] & 0x7FFF_FFFF` computes `1392955331` (BE) / `1137641043` (LE). The declared literal is implemented as normative and pinned in a test; the formula assertion is deliberately omitted. Note: the constant is a spec placeholder — re-pinning it later orphans every wallet derived under it, so it should be finalized before mainnet.

## Test plan
- [x] `cargo test -p zolana-keypair --lib` — 10 passed: SLIP-0010 nist256p1 vectors 1+2, derivation-retry vector (`m/28578H/33941`), seed-retry vector, path determinism/distinctness, coin-type pin, `from_seed` reproducibility
- [x] `cargo build -p zolana-client` (no features) and `--features "indexer-api solana-rpc"`
- [x] `cargo clippy -p zolana-keypair --all-targets -- -D warnings`; `cargo clippy -p zolana-client --all-targets --features "indexer-api solana-rpc" -- -D warnings`
- [x] `cargo test -p zolana-client --lib --features "indexer-api solana-rpc"` — 56 passed; doc-tests pass
- [x] `cargo +nightly fmt --all -- --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9742,6 +9742,7 @@ dependencies = [
  "ed25519-dalek 2.2.0",
  "hex",
  "hkdf",
+ "hmac 0.12.1",
  "p256",
  "rand 0.8.6",
  "sha2 0.10.9",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,6 +182,7 @@ ed25519-dalek = "2.2.0"
 aes = { version = "0.8", default-features = false }
 ctr = { version = "0.9", default-features = false }
 hkdf = "0.12"
+hmac = "0.12"
 zeroize = "1"
 criterion = "0.5"
 proptest = "1"
--- a/sdk-libs/client/src/lib.rs
+++ b/sdk-libs/client/src/lib.rs
@@ -48,9 +48,9 @@ pub use rpc::{
 #[cfg(feature = "solana-rpc")]
 pub use solana_rpc::{ConfirmedInstructionGroups, SolanaRpc};
 pub use user_registry::{
-    decode_user_record_account, ensure_registered, fetch_user_record_checked,
-    fetch_user_record_optional_checked, resolve_registered_address, resolved_address_from_record,
-    try_resolve_registered_address, validate_registered_keypair,
+    create_private_wallet, decode_user_record_account, ensure_registered,
+    fetch_user_record_checked, fetch_user_record_optional_checked, resolve_registered_address,
+    resolved_address_from_record, try_resolve_registered_address, validate_registered_keypair,
 };
 pub use wallet_authority::{
     AnonymousRecipientSlot, ApprovalRequest, ConfidentialRecipientSlot, EncryptedTransfer,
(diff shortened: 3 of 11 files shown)


sergeytimoshin:
Maintainer decision (2026-07-15): `TSPP_COIN_TYPE` is **formula-normative** — the recipe `SHA-256("luminous.TSPP.v1")[0..4]` big-endian `& 0x7FFF_FFFF` is authoritative, which evaluates to **1392955331**. The literal `1445561917` in this PR's `slip10.rs` was a spec placeholder that does not match the formula; #134 corrects the spec literal and pins the derived value with a provenance test.

When this rebases onto #134 (which already ships the SLIP-0010-P256 module + the ed25519 `wallet_seed` bridge), please: (1) drop the duplicate `slip10.rs` in favor of #134's, and (2) adopt `1392955331`. Keeping this PR's unique value — `ShieldedKeypair::from_seed` + `create_private_wallet` — on top. The two coin-type values derive different keys for every wallet, so they must agree before either ships a wallet.

PR #114, opened 2026-07-07
feat(client): deposit send_and_sync, asset-id lookup/registration, registry-refresh contract

tilo-14:
## Summary

Three SDK DX additions that let the examples (and integrators) drop their local helper crate, plus benchmark-based CU limits. Stacked on the wallet-onboarding PR (`main` ← #108 ← #109 ← #112 ← this).

**1. `Deposit::send` + `Deposit::wait_until_synced`** — `send` is the deposit default. The recipient-side read-your-write step is a separate `wait_until_synced(wallet, indexer, signature)`: it polls `sync_wallet` until the deposited UTXO is visible (120s/500ms), for callers that spend or display the balance immediately. Its docs state the precondition the previously-planned fused `send_and_sync` hid: the wallet must be the deposit **recipient's** — a non-recipient wallet can never see the UTXO, so the wait times out with `DepositNotIndexed` (which carries the confirmed signature) despite a successful deposit. For third-party deposits there is nothing to wait for on the sender side.

**Also:** deposits are sent with a 40k CU ceiling (benchmarked ~34k) and `DEFAULT_TRANSACT_CU_LIMIT` drops 1.4M → 200k (benchmarked ~148–152k, `program-tests/shielded-pool/CU_BENCHMARK.md`); `CreateTransfer.recipient_owner` is renamed to `recipient`, matching `CreateWithdrawal`.

**2. `fetch_asset_id` + `register_spl_interface`** — the production asset-id path and the devnet registration tool, split:
- `fetch_asset_id(rpc, mint)` reads the canonical per-mint Asset registry PDA — works on every deployment (spec: the PDA is the canonical `asset_id` source; SOL is id 1 with no PDA). New named error `AssetNotRegistered { mint }`.
- `register_spl_interface(rpc, authority, mint)` creates the interface and returns the id, idempotently — existing registrations short-circuit, and a lost registration race re-fetches instead of erroring. **Gating documented per spec:** `create_spl_interface` requires `protocol_config.protocol_authority` unless the deployment sets `spl_interface_creation_is_permissionless` (on-chain rejection: `UnauthorizedCaller`, 7003); the one-time asset counter bootstrap is *always* authority-gated. This kills the examples' hardcoded `asset_id = 2`.

**3. Registry-refresh contract** — `refresh_registry_from_chain` is now public and the behavior is documented: the sync-time backfill needs `get_program_accounts`, which the stock `ZolanaIndexer` does not implement — **with Photon as the sync source the backfill always soft-misses.** The public refresh lets callers backfill eagerly through a capable backend (e.g. `SolanaRpc`); the guaranteed per-mint path is `fetch_asset_id` + `registry.insert`. No behavior change.

All three APIs are ungated (compile with no features) and ship `no_run` rustdoc examples.

## Deferred (from the DX task list)

- **A4** (env-based devnet constructor, tree constant, `assert_deployed`): blocked on reconciling the `ZolanaClient` redesign on `feat/sdk-submit-api` with the preset work in #109.
- **A5 — sponsor-paid registration (highest-leverage embedded-wallet follow-up):** investigated — the user-registry `register` instruction has 3 accounts and the owner is the *only* signer and rent payer, so apps must pre-fund users today. Enabling a separate sponsor requires a program + interface change (add a payer account); deserves its own PR with protocol review.
- **A6**: `zolana-test-utils` is `publish = false` and SDKs must not depend on it — examples will vendor labeled scaffolding.
- **A7** (prover trust-model docs) and the examples migration (B1–B5), incl. the devnet e2e verification of the registry backfill.

## Test plan
- [x] `cargo +nightly fmt --all -- --check`
- [x] `cargo build -p zolana-client` (no features) and `--features "indexer-api solana-rpc"`
- [x] `cargo clippy -p zolana-client --all-targets --features "indexer-api solana-rpc" -- -D warnings` + `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p zolana-client --lib --features "indexer-api solana-rpc"` — 60 passed (4 new `spl_interface` tests)
- [x] doc-tests with both features (3) and featureless (2); `cargo test -p zolana-keypair --lib` — 10 passed

### DIFF
--- a/cli/src/wallet_cli/transaction.rs
+++ b/cli/src/wallet_cli/transaction.rs
@@ -30,7 +30,7 @@ pub(super) fn run_transfer(opts: TransferOptions) -> Result<()> {
         authority: &ctx.material,
         owner_pubkey: ctx.material.owner_pubkey(),
         payer: Address::new_from_array(ctx.material.funding.pubkey().to_bytes()),
-        recipient_owner,
+        recipient: recipient_owner,
         asset,
         amount: opts.amount,
     })?;
--- a/sdk-libs/client/src/actions/deposit.rs
+++ b/sdk-libs/client/src/actions/deposit.rs
@@ -1,5 +1,10 @@
 //! Proofless shield action.
 
+use std::{
+    thread::sleep,
+    time::{Duration, Instant},
+};
+
 use solana_address::Address;
 use solana_instruction::Instruction;
 use solana_keypair::Keypair;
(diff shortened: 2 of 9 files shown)


tilo-14:
Superseded by #117 (consolidated).

PR #115, opened 2026-07-07
refactor(client): rename CreateTransfer.recipient_owner to recipient

tilo-14:
## Why

The caller passes the recipient's Solana address and the SDK decides the delivery path itself (registered private wallet -> private transfer; otherwise -> public withdrawal to that address). Resolving the address to the owner's shielded address is internal mechanism, so the field should be named by caller intent. This also aligns the surface with `CreateWithdrawal.recipient`, which already takes the same concept under that name.

## What changed

- `CreateTransfer.recipient_owner` -> `recipient` (struct field and all uses in `create_transfer`, including the withdrawal fallback and approval summary).
- CLI `transfer` command updated, local renamed to keep field shorthand.
- Internal `owner` vocabulary (`ResolvedAddress.owner`, `owner_pubkey`) is untouched.

## Breaking

Struct literals constructing `CreateTransfer` must rename the field. No behavior change.

## Verification

- `cargo check --workspace` clean
- `cargo +nightly fmt --all -- --check` clean


tilo-14:
Superseded by #117 (consolidated).

PR #123, opened 2026-07-10
feat(keypair): encode shielded addresses

sergeytimoshin:
## Summary

- define a fixed 99-byte `ShieldedAddress` binary layout
- add versioned Base58Check `Display` and `FromStr` implementations
- validate embedded keys, checksum, version, and length, with a pinned golden vector

## Why

Private recipients need a self-contained printable representation before CLI and SDK flows can stop relying on mutable registry lookup. This PR only defines the codec; it does not change wallet or transfer behavior.

## Validation

- `cargo test -p zolana-keypair`
- `cargo clippy -p zolana-keypair --all-targets -- -D warnings`


sergeytimoshin:
Superseded by #133, which collapses the shielded-address codec and the private-wallet-flow changes into one PR (two commits) and restores pay-by-Solana-pubkey recipient resolution.

PR #124, opened 2026-07-10
refactor(cli): simplify private wallet flow

sergeytimoshin:
## Summary

- replace network-coupled `wallet init` with offline `wallet new` and `wallet address`
- accept self-contained shielded addresses for deposits and private transfers, with public settlement kept behind explicit `wallet withdraw`
- follow Solana CLI conventions for `-k/--keypair`, `-C/--config`, `id.json`, and importing a funding keypair
- make wallet creation exclusive and private (`create_new`, mode `0600`) and remove the redundant public `wallet sync` command
- keep the user registry optional, using it only when the existing merge-consent command needs it

## Why

The CLI should have one local wallet-selection model and one unambiguous recipient model. Creating a wallet should not require RPC access or registration, and `transfer` should never silently turn an invalid private recipient into a public withdrawal.

This is stacked on #123, which defines the printable shielded-address codec. It deliberately excludes multi-input selection, split/consolidation, and operator-command cleanup.

## Validation

- `cargo test -p zolana-client -p zolana-cli`
- `cargo clippy -p zolana-cli -p zolana-client -- -D warnings`
- `cargo test -p shielded-pool-tests --features localnet --test localnet_wallet_cli_e2e --no-run`
- manual `-C`/new/address/funding-import/permissions/no-overwrite smoke test

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -24,36 +24,42 @@ Commands:
 
 ```bash
 zolana config get
-zolana config set --keypair ~/.config/zolana/pid.json --rpc-url http://127.0.0.1:8899 --indexer-url http://127.0.0.1:8784 --prover-url http://127.0.0.1:3001
-zolana wallet init --airdrop-lamports 1000000000
+zolana config set --keypair ~/.config/zolana/id.json --rpc-url http://127.0.0.1:8899 --indexer-url http://127.0.0.1:8784 --prover-url http://127.0.0.1:3001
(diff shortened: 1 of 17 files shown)


sergeytimoshin:
Superseded by #133, which collapses the shielded-address codec and the private-wallet-flow changes into one PR (two commits) and restores pay-by-Solana-pubkey recipient resolution.

PR #125, opened 2026-07-10
feat(client): add packet-safe multi-input spending

sergeytimoshin:
## Summary

- select up to five private input notes largest-first, with `wallet utxos` and `transfer --input` for explicit selection
- keep enabled transfer shapes packet-safe and cap SPL withdrawals at three inputs
- move private proof/submission into the SDK with payer, settlement, proof-order, and tree validation
- confirm writes by committed output hash and distinguish failed, pending, confirmed, and unindexed signatures
- pin the Solana packet boundaries and prove real 2-5 input witnesses

## Why

Fragmented balances need multi-input spending, but every accepted shape must fit Solana's 1232-byte packet limit. The SDK also needs to validate that the funding key, public settlement accounts, and indexer proofs match the signed transaction before paying prover or RPC costs.

This is stacked on #124. It deliberately excludes split and merge/consolidation; those use separate transaction constructions and follow in later PRs.

## Validation

- `cargo test -p zolana-transaction -p zolana-client -p zolana-cli`
- strict Clippy across all three packages and targets with `-D warnings`
- `cargo test -p shielded-pool-tests --features localnet --test localnet_wallet_cli_e2e --no-run`
- full-message packet tests: private shapes through `{5,3}`, SPL withdrawal 3/4 boundary, SOL withdrawal at 5

The validator-backed wallet e2e was compiled but not run for this slice.


sergeytimoshin:
Superseded by #134, which collapses this stack onto current main (post-#119/#132) and layers the ed25519-default wallet identity model on top.

PR #128, opened 2026-07-12
refactor(cli): simplify operator workflows

sergeytimoshin:
## Summary
- move `create-tree` and `test-mint` to top-level operator commands while retaining wallet/RPC overrides
- define the deployed default tree once in the protocol interface and use flag -> config -> protocol-default resolution
- preflight protocol policy and account state before side effects; persist new tree keys in standard Solana format
- derive SPL associated token accounts by default while preserving explicit legacy source-account overrides
- add `config unset`, safe asset/config updates, and stricter protocol/tree account parsing
- delete the unreferenced pingpong shell scripts and cover consolidation, splitting, SOL, and SPL flows in the Rust localnet e2e

## Validation
- `cargo fmt --all -- --check`
- `cargo test -p zolana-cli -p zolana-interface`
- `cargo clippy -p zolana-cli -p zolana-interface -p xtask --all-targets -- -D warnings`
- `cargo clippy -p shielded-pool-tests --features localnet --test localnet_wallet_cli_e2e -- -D warnings`
- `cargo test -p shielded-pool-tests --features localnet --test localnet_wallet_cli_e2e --no-run`
- full runtime recipe attempted after successful SBF/prover/CLI builds; the installed Photon binary stopped the existing first e2e with JSON-RPC `Method not found` for `get_encrypted_utxos_by_tags`

Stacked on #127. Supersedes the operator/e2e portion of #98.

PR #129, opened 2026-07-12
refactor: integrate Photon into the Zolana workspace

sergeytimoshin:
## Summary

- import `helius-labs/photon-privacy` at `756e89347bb7943abbddfca4cad6c4f8a93c8894` with its full commit history, relocated under `services/photon`
- make Photon and the new `zolana-indexer-api` contract real root Cargo workspace members with one lockfile, Rust 1.95 toolchain, and shared Apache-2.0 metadata
- replace duplicated/generated indexer DTOs with shared request/response types and canonical JSON-RPC method markers used directly by Photon, `zolana-api`, and the SDK client
- build Photon from the same source revision for localnet tests instead of downloading a separately pinned release
- consolidate Photon tests, PostgreSQL migration checks, OpenAPI drift checks, Docker validation, and production image releases into root workflows
- remove standalone-repository lockfiles, workflows, code generators, publish scripts, sibling-checkout assumptions, compatibility re-exports, and obsolete dependency generations

## Stack

This is stacked on #128 (`agent/pr98-operator-e2e`). The Zolana/Rings rename remains deferred to #121.

The PR intentionally contains Photon's imported ancestry rather than a squashed source dump. `services/photon/IMPORT.md` records the source and relocation commits.

## Design notes

Photon is source-integrated but remains a separate process and independently approved deployable. The release workflow accepts version tags matching `services/photon/Cargo.toml` or explicit manual tags, pins base images and external actions by digest/SHA, smoke-tests every packaged binary as a non-root image, uses GitHub OIDC for AWS access, serializes production publication, and refuses `latest` or existing release tags.

ECR Public has no server-side tag immutability. The workflow narrows that limitation by publishing the commit tag first, retagging that exact remote image, and verifying both remote tags resolve to the same digest. Production access must remain exclusive to the workflow's OIDC role.

`photon-indexer` is `publish = false` to avoid colliding with the unrelated public crate. `zolana-indexer-api` is the sole owner of shared wire types, bounded pagination, and RPC method metadata; Photon no longer exposes local compatibility modules for those types. OpenAPI is derived from the same `RpcMethod` markers rather than manually pairing names and schemas.

Photon's direct dependencies use workspace versions. Its typed JSON-RPC adapter uses the workspace Reqwest for the six Solana methods Photon needs, and its GCS adapter uses the same client plus one cached credential provider for reads, writes, lists, and deletes. Yellowstone subscriptions use the generated Tonic client directly. Removing `cloud-storage`, `solana-client`, `solana-rpc-client`, and `yellowstone-grpc-client` eliminated the older HTTP graph, TPU/QUIC/pubsub clients, Tower 0.4, Solana pubkey 3, and OpenSSL from the Photon package graph.

The imported dependency graph is intentional for Photon's database, GCS, JSON-RPC, and Yellowstone responsibilities. Direct dependencies pass `cargo machete`; SQLx default features are disabled so the workspace does not enable its unused `any`, `json`, `macros`, or `migrate` surfaces.

The workspace repository metadata intentionally identifies this monorepo as the canonical source for publishable `program-libs`. This PR does not add crate-publishing automation, and crates that were already `publish = false` remain private.

The obsolete `program-tests/byte-compat` crate is deleted. It tested compatibility with an upstream Light parser that Photon no longer uses; both the program and integrated Photon now consume the workspace-owned tree layout. This also removes Borsh 0.10 and the old upstream Light account crates from the workspace lockfile.

The root `rust-toolchain.toml` is the only Rust version/component source for local development and CI. Nightly-only rustfmt settings that stable ignored were removed.

## Validation

- `cargo metadata --locked --format-version 1 --no-deps`
- `cargo check --locked --workspace --all-targets`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `cargo machete --skip-target-dir`
- `cargo test --locked -p zolana-indexer-api -p zolana-api -p photon-indexer --all-targets`
  - 51 Photon unit tests, 20 Photon integration tests, 4 transport tests, 4 shared-contract tests
- `cargo test --locked -p zolana-client --features indexer-api --lib indexer` (8 tests)
- `cargo test --locked -p zolana-cli` (50 tests)
- `cargo test --locked -p shielded-pool-tests --features localnet --test localnet_photon_e2e --no-run`
- `just build-programs`, explicitly using platform-tools v1.54, builds all three SBF programs
- release builds for every Photon binary, including `photon`, migrations, snapshots, and OpenAPI generation
- Photon SQLite migration `up -> status -> down --num 1 -> up`
- Photon OpenAPI generation is idempotent and validates with `swagger-cli`
- `actionlint` v1.7.7 passes for the revised workflows
- pinned Docker base-image digests and action SHAs resolve

A local Docker daemon and PostgreSQL server were unavailable, so the actual container build/smoke and PostgreSQL-specific migration cycle are delegated to the new CI jobs.

## Operational follow-ups

- configure and protect the `photon-production` GitHub environment
- define its `ECR_HELIUS_PROD_AWS_ROLE_ARN` variable with an AWS OIDC role scoped to `repo:helius-labs/zolana:environment:photon-production` and the Photon ECR Public repository
- restrict that environment's deployment tags to `photon-v*` and require approval for manual releases
- migrate external deployments from the standalone workflow's `latest` / `<run>-<sha>` tags to this workflow's version or `sha-*` tags before archiving the old repository
- retain/archive the old repository for branch, PR, and operational history not reachable from its imported `main`
- generate a full transitive dependency license/SBOM inventory separately if production distribution policy requires it


services/photon/tests/data/transactions/rings_e2e/3JAHH578NVLSh6Z3x2tXnNV4U9digpQb5CsFLrfac6fRNRCNQaRVsWNNrdQUT8PzUEw2MH78MkuKCZxqDoLKqNcX
ananas-block, line 1: should we even create these transactions? we can just run photon on a test validator test

PR #133, opened 2026-07-13
feat: shielded address codec + pay-by-pubkey wallet flow

sergeytimoshin:
Supersedes and collapses #123 and #124.

## Commits
1. `feat(keypair): encode shielded addresses` — versioned Base58Check codec for the full `ShieldedAddress` (signing / nullifier / viewing keys). Required by the confidential recipient slot, which is tagged by and proof-bound to the recipient's signing pubkey (see `docs/spec.md` `RecipientSlot` / default-zone tagging).
2. `feat(cli): shielded-address and pay-by-pubkey wallet flow` — the wallet CLI on a self-contained shielded identity, with `transfer`/`deposit --to` accepting a shielded address (direct) or a Solana pubkey (user-registry lookup); an unregistered pubkey transfer silently falls back to a public withdrawal, a deposit to one errors.

## Why the full address (not the compressed `(owner_hash, viewing_pk)` form)
A default-zone confidential transfer writes the recipient's signing pubkey into the on-chain recipient slot and derives the output owner from it, so the sender needs the full address; `(owner_hash, viewing_pk)` is insufficient. The signing pubkey is published on-chain in every confidential transfer to that recipient anyway, so the address discloses nothing new. The compressed form remains the right primitive for anonymous policy-zone transfers (a separate feature).

## Validation
- `cargo test -p zolana-cli` / `-p zolana-client`; `cargo clippy -- -D warnings`; `cargo fmt --check`.
- Localnet wallet-cli e2e (`just test-localnet-e2e-photon`) deploys the user-registry program and covers the SOL+SPL cycle, direct shielded-address transfers, the public-withdrawal fallback, and registered pay-by-pubkey.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1027,6 +1027,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bf88ba1141d185c399bee5288d850d63b8369520c1eafc32a0430b5b6c287bf4"
 dependencies = [
+ "sha2 0.10.9",
  "tinyvec",
 ]
 
@@ -9737,6 +9738,7 @@ name = "zolana-keypair"
 version = "0.1.0"
 dependencies = [
  "aes",
+ "bs58",
  "ctr",
  "cucumber",
  "ed25519-dalek 2.2.0",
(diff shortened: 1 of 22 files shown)


sergeytimoshin:
Superseded by #134, which collapses this stack onto current main (post-#119/#132) and layers the ed25519-default wallet identity model on top.

PR #135, opened 2026-07-15
chore: sdk changes for turnkey poc 

swenhelius:


swenhelius:
Closing this as obsolete for the Turnkey POC. Current main has most of what we need. The remaining changes here are broader hardening and should be proposed separately if at all.

PR #139, opened 2026-07-16
Reject transact with both public amounts (C-01)

Atamanov:
A transact that set both public_sol_amount and public_spl_amount settled only the SPL leg - the account parser picks a single settlement branch while the proven SOL leg never moved, minting an unbacked SOL note.

### DIFF
--- a/program-tests/shielded-pool/tests/transact/both_amounts.rs
+++ b/program-tests/shielded-pool/tests/transact/both_amounts.rs
@@ -8,23 +8,46 @@
 
 #[path = "../common/setup.rs"]
 mod common;
-#[path = "../common/transact.rs"]
-mod transact_common;
 
 use solana_keypair::Keypair;
 use solana_message::Message;
 use solana_signer::Signer;
 use solana_transaction::Transaction;
 use zolana_interface::{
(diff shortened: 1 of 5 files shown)


security-reports/C-01-simultaneous-sol-spl/REPORT.md
ananas-block, line 1: pls remove from pr :)
I dont see a reason to commit this

ananas-block (commented):
Thanks! Once we remove the report lgtm

program-tests/shielded-pool/tests/transact/both_amounts.rs
```diff
@@ -0,0 +1,123 @@
...
+//! while the proven SOL leg never moved. The fix rejects both-present up front.
+//!
+//! Asserts the program returns `BothPublicAmountsSet` (7023) and moves no tokens.
+//! The reject precedes proof verification, so no real proof is needed. Skips when
+//! the program `.so` is missing.
+
+#[path = "../common/setup.rs"]
+mod common;
+#[path = "../common/transact.rs"]
+mod transact_common;
```
ananas-block: this is probably pulling in these files in a way those were not before and makes ci fail

ananas-block (approved):

PR #134, opened 2026-07-15
feat: private-wallet lifecycle — ed25519 identity, split, merge, utxos

sergeytimoshin:
1. **identity.** ed25519-default: the Solana funding key IS the shielded identity (`from_solana_keypair`; viewing key = flat HKDF); ed25519 files store the funding secret + `owner_hash` and re-derive on load; `--p256` mints an independent random P256 identity (scalars stored). `wallet new` is offline; `wallet register` is the explicit, strict on-chain step.
2. **split** (1→8). `wallet split --parts N`: a 1-input → N-output self-transfer filling `IN1_OUT8`.
3. **merge** (8→1). `wallet merge` (auto-sweep ≤8 plain utxos, or `--input` hashes);
4. **utxos.** `wallet utxos` lists spendable utxos per asset;

PR #144, opened 2026-07-17
fix: hide split output count (zero-value self-output pads)

sergeytimoshin:
The split leaked its real output count N: the N value-bearing outputs shared the owner view tag while the (8-N) padding slots were `owner=0` dummies with random tags, so an observer counted N and learned the split size.

**Fix (client-only, no circuit change):** the padding slots are now **real zero-value self-outputs** (owner-bound, amount 0, blinding derived from the shared seed) instead of owner=0 dummies. Every slot is a genuine owner-bound commitment publishing the owner's view tag, so:
- all 8 published tags are identical -> N is hidden;
- the circuit's per-output owner binding (`Poseidon(owner_pk, nullifier_pk) == owner_hash`) holds for every slot;
- balance still holds (`N*per_output + zeros = input`).


sergeytimoshin:
Subsumed by #134. The #2 fix (hide split output count via zero-value self-output pads) is already in #134 as commit 0139814a, refined further by 755e0309 ("close remaining review findings"), after #134 was rebased onto main. #134 is now the single PR carrying the whole private-wallet stack including this fix. Closing to avoid a conflicting/empty duplicate.

PR #154, opened 2026-07-21
feat(sdk-tests): confidential RFQ settlement example, no escrow

ananas-block:
## What

Adds `sdk-tests/rfq`: a confidential RFQ settlement where a maker and taker co-sign a single shielded-pool `transact` that swaps SOL for USDC with **no escrow and no custom program**.

The mechanism is one raw `transact` with two ed25519 co-signers: the taker input's `eddsa_signer_index` is overridden to point at the taker signer account appended to the instruction, and the whole thing is sent as a v0 transaction with an address lookup table (IN2_OUT2 fits under the 1232-byte packet limit).

## Contents

- `tests/rfq.rs` + `tests/shared.rs` — localnet end-to-end settlement against a validator + Photon + prover + Squads smart account. Asserts both recipients' indexed UTXOs against fully-reconstructed expected values (owner, amount, blinding, empty data) and that the spent sides are zero.
- `tests/bench_cu.rs` — mollusk CU benchmark mirroring the swap tooling (`light-program-profiler`), regenerating `BENCHMARK.md` via `just bench-rfq`. The shielded-pool program is built with `profile-program` so its `#[profile]` functions appear in the per-function CU table; proving time is the warm (key-loaded) SPP transfer proof.
- `justfile` — `test-rfq-validator` and `bench-rfq` recipes.
- CI — `test-rfq-validator` added to the `test-localnet-photon` matrix (the benchmark is not run in CI, matching `bench-swap`).

## Benchmark snapshot

~155k CU total (`verify_groth16` dominating at ~93k), 116 ms warm proving, 617 B instruction data, 964 B v0+ALT transaction.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -206,6 +206,8 @@ jobs:
             recipe: test-swap-and-escrow-validator
           - name: dynamic-swap lifecycle
             recipe: test-dynamic-swap
+          - name: rfq settlement
+            recipe: test-rfq-validator
     env:
       AGAVE_VERSION: v4.0.2
       # GH_TOKEN is for the gh-release swap/escrow key downloads (ensure-swap-keys /
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5425,6 +5425,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "rfq-test"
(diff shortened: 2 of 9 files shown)


tilo-14 (approved):
lgtm

PR #160, opened 2026-07-25
fix(merge): bind merge_transact to the merged owner's own registry record

tilo-14:
## Bug

- `merge_transact` takes the merged owner's signing key, viewing key, and `merging_enabled` flag from a `user_record` account, checking only that the account is owned by the registry program and starts with the `UserRecord` discriminator. Any record the registry created passes both checks, so the caller chooses which record the pool reads.
- The proof covers one of those three fields. The circuit checks `signing_pk_field` against the owner hash in each input and output UTXO, so a substituted signing key fails verification. The viewing key is only checked against the ciphertext, and `merging_enabled` is not a circuit input.
- A caller holding the owner's secrets (a current or former sync delegate) can therefore supply another record and have the output encrypted with a key the owner does not hold. The owner cannot decrypt it, so the output's blinding stays unknown to them and they cannot compute its nullifier, while the inputs are nullified either way. The merged balance is unspendable. The same substitution reads `merging_enabled` off the chosen record, so turning the flag off does not stop the merge.

## Fix

- The shielded pool re-derives the record's canonical PDA from the `owner` and `bump` the record stores, and returns `InvalidUserRecord` on a mismatch. The address is then what identifies the owner: on the ed25519 rail the signing key comes from that same `owner`, which the circuit already checks against the UTXOs.
- That is not enough on the P256 rail. The address establishes whose record it is, but `register` writes `owner_p256` without checking that the registrant holds the matching private key, so two records could hold the same key. `register` and `update_keys` now record the identity in a `P256OwnerClaim` PDA and reject a key that another record already holds, or whose x-coordinate is a registered owner's address. The second case is the parity collision: owner identity drops the SEC1 prefix byte, so `0x02 || solana_address` produces the identity of that Solana address.
- Claims are first-come rather than proof of possession, so an identity claimed before its holder registers stays with the first claimant. Closing that needs a secp256r1 signature at registration, which cannot land here because the registry tests register placeholder P256 keys with no known private key.

## Test plan

- [x] `just build-programs`, then `cargo test -p shielded-pool-tests --test merge_user_record`: 3 passed (3 failed before the fix, `Custom(7008)` where `Custom(7018)` is required)
- [x] `just test-shielded-pool`: green, including the prover-backed p256 transact test and 33 BDD scenarios
- [x] `cargo test -p user-registry-tests`: 29 scenarios, 5 of them new
- [x] localnet merge lifecycle, both rails with real proofs and registrations (`merge.feature`, `merge_eddsa.feature`): 18 scenarios passed
- [x] `cargo test -p zolana-wallet`, `cargo fmt --check`, `cargo clippy` on the touched crates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6210,6 +6210,7 @@ dependencies = [
  "serde_json",
  "serial_test",
  "solana-account 2.2.1",
+ "solana-account 3.4.0",
  "solana-address 2.6.1",
  "solana-compute-budget-interface",
  "solana-instruction 2.3.3",
@@ -6233,6 +6234,7 @@ dependencies = [
  "zolana-test-utils",
  "zolana-transaction",
  "zolana-tree",
+ "zolana-user-registry-interface",
  "zolana-wallet",
 ]
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1639,7 +1639,7 @@ struct ZoneDepositIxData {
 | --- | --- | --- | --- | --- |
 | 1 | tree_account | x |   | nullifier queue + nullifier tree + UTXO tree |
 | 2 | payer |   | x | fee payer; any account may run the merge |
-| 3 | user_record |   |   | read-only; the owner's [registry](#registry) record. SPP checks `merging_enabled == true` and binds the proof's signing / viewing `pk_field`s to it |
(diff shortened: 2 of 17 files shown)


tilo-14:
One deployment note the PR body does not cover: records that already carry an
`owner_p256` when this ships have no claim account, so the exclusivity half is
void for them until someone claims their identity. A new registrant could take
it, since no claim exists to refuse them. The migration is one `update_keys`
call per existing P256 record with the same key it already holds, which creates
the claim (the re-claim path is idempotent for the record's own owner). For a
deployment with a non-trivial number of live records, that argues for doing the
claim creation as part of the upgrade rather than leaving it to wallets.

PR #161, opened 2026-07-25
feat: reimburse foresters for nullifier batches

ananas-block:
**Changes:**
1. add reimbursement for forester batch Merkle tree update
2. add collection of lamports to fund forester reimbursement at nullifier queue insertion. Lamports must scale dynamically based on the number of nullifiers and be paid to the tree account the nullifiers are inserted into.
3. use 1 sol transfer via cpi and distribute other lamports by manipulating lamports balances of tree accounts directly without cpi.


tilo-14 (approved):
lgmt

PR #165, opened 2026-07-26
test: explicit Rust suites, invariants ledger, and PR171-protocol coverage

sergeytimoshin:
Replaces the BDD/cucumber suites with explicit Rust tests and brings the test surface to the post-PR171 protocol.

Four pieces were extracted for independent review: #180 (dependency move, land first), #179 (swap SDK account-layout fix), #178 (Go circuit tests). The `smart-account-client` role table moved to `program-tests/test-utils` and stays here, since only test code consumes it.

## Tests

- Domain-first binaries (`admin/`, `deposit/`, `transact/`, `merge/`, `spl_interface/`, `cross_cutting/`, `localnet/`) over a shared `src/support` library. 35 `[[test]]` targets, 10 feature-gated, versus 10 before. Features/steps/world files deleted.
- Invariants ledger (`program-tests/shielded-pool/invariants/`): 242 invariants — 206 covered, 3 covered on companion branches (`- [~]`, excluded from the count), 21 partial, 11 N/A post-PR164, 1 pointer. Each carries statement, severity, error code, location, and covering test. `tools/check-test-hygiene.sh` enforces tallies and citations in CI.
- Tiers: LiteSVM fast tier, Mollusk exact-rejection sweeps, Groth16 proof matrices, localnet + Photon end-to-end including the 20-batch forester leg asserting persisted-root equality, and CU benchmarks.
- `zolana-test-utils` shared localnet harness, typed `Rejection` assertions, nextest `serial-validator` group with retries scoped to validator suites.

## Non-test deltas

- `transact/interface_transfer.rs` — public-slot writes go through `.get()`/`.get_mut()` returning `InvalidTransactShape` instead of bare indexing. Behavior-preserving on every reachable input; a malformed future caller gets an error rather than a panic.
- `#[cfg(feature = "test-sbf")] pub mod testing` plus `pub(crate)` -> `pub` on `solana_pk_hash`, `public_input_hash`, and `amount_field`, so the unit tests moved out of the program crate can reach them. `test-sbf` is off by default and only `program-tests/shielded-pool` enables it, so the shipped `.so` is unaffected.
- `zolana-client` — `spawn_prover_with_artifacts`; `PublicInputs` and `hash()` made public so tests reuse the canonical public-input assembly instead of mirroring it.
- `program-tests/test-utils` — gains the smart-account `roles`/`settings` modules. Their docs record that `xtask init-protocol` derives the same five roles independently rather than sharing the table; deduplicating means driving xtask's creation sequence from `Role::ALL`, left as a follow-up because `init_protocol.rs` has no tests and a wrong role order would misassign authorities on a live cluster.

## Coverage

Every main BDD scenario is mapped and ported, most strengthened to exact error codes and rollback snapshots rather than string matches. Both pre-existing holes are closed: on-chain double-spend, and ed25519 spend-signature with real proofs. A quality pass removed ~2.4k lines of duplicated and tautological tests.

`- [x]` means an invariant has at least one covering test, not that every clause of its statement is asserted. 21 entries are marked partial; the widest is INV-XC-04 (rollback), which names the instructions still lacking account-equality assertions.

Land after #174, #175, #176, #177, and #180. The P256 rail returns via #172; the P256 N/A entries and INV-XC-32 need re-scoping then.


program-tests/shielded-pool/tests/cross_cutting/cu_budget.rs
```diff
@@ -0,0 +1,166 @@
...
+const CREATE_TREE_CU_CEILING: u64 = 1_800; // observed 581
+const PAUSE_TREE_CU_CEILING: u64 = 800; // observed 250-251
+const CREATE_ASSET_COUNTER_CU_CEILING: u64 = 14_000; // observed 4_600
+const CREATE_SPL_INTERFACE_CU_CEILING: u64 = 23_000; // observed 7_638
+const DEPOSIT_SOL_CU_CEILING: u64 = 90_000; // observed 38_393
+const DEPOSIT_SPL_CU_CEILING: u64 = 100_000; // observed 39_424
+const CREATE_ZONE_CONFIG_CU_CEILING: u64 = 20_000; // observed 6_658
+const UPDATE_ZONE_CONFIG_CU_CEILING: u64 = 450; // observed 141
+const ZONE_DEPOSIT_SOL_CU_CEILING: u64 = 120_000; // observed 47_441
+const UPDATE_ZONE_CONFIG_OWNER_CU_CEILING: u64 = 700; // observed 218
```
ananas-block, line 26: imo the ceiling asserts should be more aggressive

ananas-block: not sure its really necessary though if we just check in CI that CU_BENCHMARK.md is committed and has no diff in CI

sergeytimoshin:
Extracted from this PR for independent review (see the updated description):

- #180 — dependency move (mollusk 0.14 / litesvm 0.15 / agave 4.1) + harness migration. **Must land before this PR.**
- #179 — `zk-program-swap` TVE take account-layout fix. Independent; broken on main since #171, so worth landing early.
- #178 — Go circuit tests for the dummy-nullifier queue-poisoning class. Independent.

The `smart-account-client` role table was moved to `program-tests/test-utils` rather than extracted, since only test code consumes it.

ananas-block (dismissed):

ananas-block (approved):

PR #170, opened 2026-07-26
feat: add single user-facing TypeScript SDK

swenhelius:
- takes just the ported parts of #159 that are needed for a single `@zolana/sdk` so as to reduce LOC

- on solana kit
- incl. 1 RPC client, 
- instruction builders,
- high-level wallet actions

- no anon support
- simplified txn builders
- no P256 nor ring support implemented


sdk-libs/hasher-wasm/src/lib.rs
ananas-block, line 1: do we need this at all? imo it is ok to have 1 functional test with hardcoded outputs to check that poseidon constants used in rust and ts match

sdk-libs/ts/src/hasher/poseidon.wasm
ananas-block, line 1: pls remove

ananas-block (approved):

PR #177, opened 2026-07-29
refactor: remove legacy merge ciphertext encoding

sergeytimoshin:
## Summary

Removes `OutputDataEncoding::VerifiablyEncrypted` and `EncryptedScheme::Merge` from the event/wallet encoding.

The merge verifiable-encryption gadget was removed in PR171 (merge outputs are now ciphertext-free, derived deterministically), so nothing produces this variant; SPP has not launched, so no legitimate historical consumer exists; and PR172 (P256 restoration) does not use it. It was dead production code kept alive by test fixtures. It is the last variant, so no wire tags are renumbered.

Changes:

- `program-libs/event`: delete the variant, tag constant, encoder, and decode arms.
- `sdk-libs/transaction` wallet sync: delete the legacy decode arm.
- Photon rings fixtures: use a local `encode_encrypted` helper instead of the deleted public encoder.
- `EncryptedScheme::Merge = 6` removed; byte 6 now decodes as `BadDiscriminator`.

**Confirmation requested from reviewers**: no live cluster has ever emitted tag-2 (`VerifiablyEncrypted`) payloads. If any such data existed, wallets would previously soft-skip it and would now hard-error on decode; pre-launch, there is none.

Also included: `apply_cached_tree_updates` exits its loop with `break` + tail `Ok(event)` instead of early `return` — behavior-identical, but the CU profiler's end marker only fires on the tail path, so the function had silently vanished from benchmark reports (its row is back, ~33k CU for the 120-entry cascade).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4364,7 +4364,6 @@ dependencies = [
  "tracing-subscriber",
  "utoipa",
  "yellowstone-grpc-proto",
- "zolana-batched-merkle-tree",
  "zolana-event",
  "zolana-indexer-api",
  "zolana-interface",
@@ -5937,7 +5936,6 @@ dependencies = [
  "ark-bn254 0.5.0",
  "ark-ff 0.5.0",
  "arrayvec",
- "bincode 2.0.1",
  "borsh",
  "bytemuck",
  "groth16-solana",
@@ -5945,8 +5943,6 @@ dependencies = [
  "light-program-profiler",
  "pinocchio 0.11.2",
  "pinocchio-system",
- "serde_json",
- "solana-loader-v3-interface 8.0.1",
  "spl-token-2022-interface",
  "tinyvec",
  "zolana-account-checks",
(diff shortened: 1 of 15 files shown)


ananas-block:
please hold off on deleting we will need this for custom rings with auditor

sergeytimoshin:
Closing: main now actively pins the variant this PR removes, so the hold has become structural rather than temporary.

#165 merged with `program-libs/event/tests/output_data.rs::…` pinning the `VerifiablyEncrypted` wire shape, documented as "reserved for upcoming auditor encryption flows", plus three new `OutputDataEncoding::VerifiablyEncrypted` match arms (localnet photon encrypted_transfer + forester, and the client proving test). Rebased onto current main this branch no longer compiles, and making it compile would mean deleting that test — reversing the review decision (ananas-block: the variant is needed for custom rings with auditor) rather than implementing it.

The behavior-neutral batched-tree update fix that rode along here was not part of the variant removal and can be re-extracted on its own if still wanted.

PR #178, opened 2026-07-31
test(circuits): pin the dummy-nullifier queue-poisoning class as unsatisfiable

sergeytimoshin:
Regression tests for the attack class closed by the circuit hardening in #171. A padding/dummy input slot must not be able to publish an attacker-chosen nullifier.

The old circuits left a dummy slot's public `Nullifier` a free witness: the derived-nullifier binding was gated on `notDummy` and non-inclusion was skipped for padding. The program queues every input slot's nullifier, and batch append requires `low < value < next`, so one arbitrary or tree-resident value halts the nullifier queue permanently.

- `spp_transaction` — an attacker-chosen nullifier, nullifier `0`, and two slots sharing one nullifier each fail to solve.
- `spp_merge` — a victim's real nullifier parked in a dummy slot fails to solve, so a merge delegate cannot burn a UTXO without merging its value.
- Positive control — the derived `MergeDummyNullifier` witness solves, and the host-side derivation is pinned against the circuit's domain and argument order.

Each negative refreshes the public input hash first, so the constraint under test is the sole rejecting one. Test-only; extracted from #165. Independent of the other extracted PRs.

### DIFF
--- a/prover/server/circuits/spp_merge/dummy_nullifier_attack_test.go
+++ b/prover/server/circuits/spp_merge/dummy_nullifier_attack_test.go
@@ -1,0 +1,98 @@
+package merge_test
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/consensys/gnark-crypto/ecc"
+	"github.com/consensys/gnark/frontend"
+	"github.com/consensys/gnark/test"
+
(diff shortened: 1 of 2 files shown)


sergeytimoshin:

ananas-block (approved):

PR #172, opened 2026-07-29
refactor: restore spp p256

ananas-block:
Changes:
1. nullifier tree update - remove already cached early return to allow for easier recovery of invalid cached state
2. update zone config owner - remove new authority pubkey from instruction data we already send it as account info
3. rename `spl_asset_vault` -> `spl_interface`
4. remove `spl_asset_registry` from  deposit instruction
5. `depositor` -> `token_authority`
6. `vault` -> `spl_interface`
7. simplify owner signers in program, use a section in accounts array instead of indices
8. simplify owner signers in circuit, include payer in signers array
9. add `CircuitId::ZoneP256`
10. add zone transfer VKs
11. registry program, require p256 signature with transaction introspection for registry account creation
12. `SPL_ASSET_VAULT_PDA_SEED` -> `SPL_INTERFACE_PDA_SEED`

PR #179, opened 2026-07-31
fix(swap-sdk): carry both tree accounts in the verifiable-encryption take

sergeytimoshin:
#171 split the shielded-pool transact account list into separate input and output tree slots. The derived take rail was updated with it; the verifiable-encryption take rail was not.

The swap program is unchanged and forwards these accounts into the SPP transact CPI, so every `TakeVerifiableEncryption` this SDK builds is malformed against the current program: the CPI reads the system program where it expects the output tree, and `ORDER_AUTHORITY_SIGNER_INDEX` selects the wrong account for the `input_owner_pk_hash` match.

- Adds the missing tree meta and moves the signer index 3 -> 4, matching `take/instruction.rs` exactly.
- Uncaught because this rail has no end-to-end coverage on main, only a circuit test that never builds the instruction. Behavioral coverage arrives with #165.

Extracted from #165. Independent of the other extracted PRs.

### DIFF
--- a/sdk-tests/zk-program-swap/sdk/src/instructions/take_verifiable_encryption/instruction.rs
+++ b/sdk-tests/zk-program-swap/sdk/src/instructions/take_verifiable_encryption/instruction.rs
@@ -16,12 +16,12 @@ pub struct TakeVerifiableEncryption {
 }
 
 /// The order utxo (input 0) is owned by the order-authority PDA appended readonly
-/// after `tree`; the swap program signs for it via `invoke_signed`. The taker
-/// input is signed by the SPP payer (account index 0). The signer index
+/// after the system program; the swap program signs for it via `invoke_signed`.
+/// The taker input is signed by the SPP payer (account index 0). The signer index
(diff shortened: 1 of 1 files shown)


sergeytimoshin:

ananas-block (approved):

sergeytimoshin:
Obsolete: both halves of this fix are resolved on main by #172.

1. **Account list** — main's `take_verifiable_encryption/instruction.rs` already carries `tree` twice and is now byte-identical to the derived `take` rail:
```
payer, payer, tree, tree, SHIELDED_POOL_PROGRAM_ID, Pubkey::default(), order_authority_pda()
```
2. **Signer index** — the other half sets `spp_proof.inputs[0].eddsa_signer_index`, but #172 removed `eddsa_signer_index` from `InputUtxo` entirely (authorization now comes from the payer-first owner-signer run in the accounts array). `InputUtxo` is now just `nullifier_hash`, `nullifier_tree_root_index`, `utxo_tree_root_index`, so this cannot compile.

Nothing in the diff still applies. The regression coverage referenced in the commit message still lands with #165.

PR #180, opened 2026-07-31
build: unify on mollusk 0.14, litesvm 0.15, and the agave 4.1 crate line

sergeytimoshin:
Dependency prerequisite extracted from #165 so the graph move is reviewable apart from the test rewrite that motivated it. The scenario suites are untouched and still green.

The set moves together and cannot be split further: the profiler is pinned to its mollusk-0.14 branch, so any crate left on mollusk 0.3 puts two mollusk majors in one graph and fails with `expected mollusk_svm::Mollusk, found Mollusk`; mollusk 0.14 needs the agave 4.1 solana crates; and litesvm 0.15's typed `FailedTransactionMetadata` is what `zolana-program-test`'s `Rejection` surface is built on.

- `mollusk-svm` 0.3 -> 0.14, declared once in the workspace root; the per-crate `mollusk-solana-{account,instruction,pubkey}` 2.2 aliases are gone.
- `litesvm` 0.12 -> 0.15, `solana-sdk` 3 -> 4, `solana-account` 3.4 -> 4.3, `solana-message` 3.1 -> 4.2, `spl-token-2022` 10 -> 11, `anchor` 1.0.2 -> 1.1.
- Pins against crate duplication: `solana-hash =4.5.0`, `solana-reward-info =6.2.0`, `solana-transaction =4.1.5` (4.2 wants `solana-hash` 4.6). The `solana-transaction` pin is new here; #165 holds 4.1.5 in its lockfile without a manifest constraint, so decide which is right.
- `solana-transaction`'s `bincode` feature became an optional dependency in 4.x; the five crates selecting it now select `serde`.
- mollusk 0.14 API across the seven `bench_cu` harnesses: `add_program` drops the loader argument, `add_program_with_elf_and_loader` is now `add_program_with_loader_and_elf(id, loader, elf)`.
- `zolana-program-test` gains `Rejection`, `TransactionTrace`, and `ProgramTestError::TransactionFailure`. `Rejection::assert_client` needs `ClientError::SolanaRpcTransaction`, so the `SolanaRpc` send paths return that typed variant instead of a formatted string.
- `zolana-tree` exports `POOL_UTXO_HEIGHT` so `bench/tree` drops the duplicated literal.
- `cucumber` stays until #165 removes the scenario suites.

Verified: fmt, clippy `-D warnings`, and `check --workspace --all-targets` clean; the existing shielded-pool scenario suite passes unchanged (44 scenarios, 197 steps), confirming litesvm 0.15's default blockhash check does not affect it.

Land before #165. Rebasing #165 onto this conflicts in the seven `bench_cu` files and `Cargo.lock` — take #165's version, which rewrites them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,143 +56,49 @@ dependencies = [
 
 [[package]]
 name = "agave-feature-set"
-version = "2.3.13"
+version = "4.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a2c365c0245cbb8959de725fc2b44c754b673fdf34c9a7f9d4a25c35a7bf1"
+checksum = "d4790979fc2a3c1c494c9cb84e8b514da4b8c6a992a460a077b31b07657606f8"
 dependencies = [
  "ahash 0.8.12",
- "solana-epoch-schedule 2.2.1",
- "solana-hash 2.3.0",
- "solana-pubkey 2.4.0",
- "solana-sha256-hasher 2.3.0",
- "solana-svm-feature-set 2.3.13",
-]
-
-[[package]]
-name = "agave-feature-set"
-version = "3.1.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bfe79fc4c114c51ea8461d829bb49853a21a76c7c8ef20e9041b071558f628ce"
-dependencies = [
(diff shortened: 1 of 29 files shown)


sergeytimoshin:

ananas-block (dismissed):

ananas-block (approved):

PR #182, opened 2026-08-01
feat(ts-sdk): add instruction-level client example + api surface refactor

tilo-14:
The TypeScript example follows the Rust deposit, transfer, and withdrawal flow. These changes add the client behavior it needs and align derived transaction inputs with Rust.

## Photon
- **Endpoint resolution:** Moves service URL selection into one helper. Photon and the prover reuse `solanaRpcUrl`; empty configurations use local ports.
- **Query parameters:** Keeps URL parameters such as `api-key` when adding a Photon method path.
- **Indexed transaction confirmation:** Adds `confirmPrivateTransaction()` to wait for Solana confirmation and Photon indexing, matching the Rust example before it reads transaction outputs.

## SDK changes
- **Solana signer:** Adds an Ed25519 Solana signer adapter for shielded keypairs, matching Rust and rejecting P256 keys with the corresponding `NotEd25519` error.
- **Balance decryption:** Adds a helper to decrypt indexed outputs into private balances, matching Rust's `decrypt_transactions`.
- **Deposit tags:** `createDeposit` tags outputs with `recipient.confidentialViewTag()` (signing-key tag) so Photon indexing and `syncWallet` can find deposits. The previous viewing-key tag made deposits invisible to wallet sync.
- **Prover assembly:** Derives the canonical SPL interface bump from the mint, matching Rust's `pda::spl_interface_bump`. TypeScript assembly is asynchronous because `@solana/kit`'s `getProgramDerivedAddress()` returns a promise.
- **Instruction naming:** Renames SPL interface and recipient-owner fields to match the Rust SDK without changing the encoded instruction layout.
- **Fee-payer inputs:** Stores the fee-payer address and derives its signer hash during prover assembly, matching Rust's `signer_pk_hashes`.


sdk-tests/client/rust/deposit_transfer_withdraw.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    // Load the funded fee payer and localnet settings, then connect.
+    let client = ZolanaClient::from_urls(SolanaRpc::new(rpc_url), &indexer_url, prover_url, tree);
+
+    // Mints that are interoperable with Solana Rings
+    // are registered with the AssetRegistry.
+    let assets = AssetRegistry::default();
+    // SPL alternatives below assume a hypothetical `spl` fixture. To use them,
+    // replace the active SOL line with its adjacent `// SPL:` lines and register:
+    // SPL: let mut assets = AssetRegistry::default();
```
tilo-14:
```suggestion
    // Toggle for SPL 
```

sdk-tests/client/typescript/deposit-transfer-withdraw.test.ts
```diff
@@ -0,0 +1,185 @@
...
+          recipientOwnerHash: senderAddress.ownerHash(),
+          blinding: randomBlinding(),
+          amount: DEPOSIT_AMOUNT,
+        },
+      ],
+    });
+
+    // 2. Send and confirm like any Solana transaction.
+    const depositSignature = await sendAndConfirm([depositIx]);
+    await client.waitForIndexedTransaction(depositSignature);
```
ananas-block: pls remove this, we should use the config in getShieldedTransactionsByTags to ensure that the indexer has indexed the latest slot or up to the timestamp of depositSignature

ananas-block: pls remove

sdk-libs/client/src/client.rs
```diff
@@ -1156,7 +1162,7 @@ async fn wait_for_rpc_confirmation_async<R: AsyncRpc>(
 /// is indexed. Matching the event against the transaction's view tags would add
 /// no guarantee and would reject legitimate transactions whose events share a
 /// tag.
-fn wait_for_indexed_transaction(
+fn poll_indexed_transaction(
```
ananas-block: why did we rename this fn to `poll_indexed_transaction` and `confirm_private_transaction` to `wait_for_indexed_transaction` shouldnt the latter also be named `poll_indexed_transaction_async`

sdk-libs/ts/src/client/prover/assembly.ts
```diff
@@ -59,31 +60,31 @@ export function circuitUtxo(value: object): CircuitUtxo {
...
 
-export function intoProver(
+export async function intoProver(
   proofInputs: SppProofInputs,
   spendProofs: readonly SpendProof[],
   dummyNullifierProofs: readonly NonInclusionProof[] = [],
-): ProverInputs {
-  return assemble(proofInputs, spendProofs, dummyNullifierProofs).proverInputs;
+): Promise<ProverInputs> {
+  return (await assemble(proofInputs, spendProofs, dummyNullifierProofs)).proverInputs;
```
ananas-block: is this really async?

ananas-block: Good catch — the `async` here is real but avoidable. The only `await` in the chain is `splInterfaceBump(transfer.mint)` when assembling SPL interface transfers, which re-derives a PDA through Kit's `getProgramDerivedAddress`. That API is Promise-only by design and won't get a sync variant (anza-xyz/kit#924, closed as not planned — Kit hashes with `SubtleCrypto.digest()` and won't bundle a userspace sha256).

But the re-derivation itself is redundant: the SPL `SettlementTransfer` already carries the derived `splTokenInterface` address, which the caller obtained asynchronously upfront. We'll carry the canonical bump alongside it (derive via `splInterfaceWithBump` at that call site), drop the `await` in assembly, and make `intoProver`/`assemble` fully synchronous.

sdk-tests/client/typescript/setup.ts
```diff
@@ -0,0 +1,174 @@
...
+  SPL_TOKEN_PROGRAM_ID,
+  ShieldedKeypair,
+  createZolanaClient,
+  initializePoseidon,
+  type Bytes32,
+} from "@zolana/sdk";
+
+// The SDK's localnet default; the airdrop is a localnet-only operation and
+// targets the same validator the bare client connects to.
+const LOCALNET_RPC_URL = "http://127.0.0.1:8899";
```
ananas-block: The example ignores the URLs `_test-ts-live` passes. This file hardcodes 8899 for the airdrop and `createZolanaClient()` connects to 8899/8784/3001, while the recipe boots the stack on `ZOLANA_PORT_OFFSET`-shifted ports and exports `ZOLANA_LOCALNET_URL` / `ZOLANA_INDEXER_URL` / `ZOLANA_PROVER_URL` (the SDK e2e reads them in `live-helpers.ts`). With a non-zero offset, `just test-ts-example` and `just test-ts-e2e` poll dead ports until the 600s timeout. CI runs offset 0 and will not catch this.

Keep the example body zero-config, but have setup.ts read the env vars and pass explicit URLs when they are set, or fail fast when an offset is configured.

sdk-tests/client/typescript/deposit-transfer-withdraw.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+import { sendAndConfirmFactory, setup } from "./setup.js";
+
+const DEPOSIT_AMOUNT = 1_000_000_000n;
+const TRANSFER_AMOUNT = 300_000_000n;
+const WITHDRAW_AMOUNT = 300_000_000n;
+
+describe("example: deposit, transfer, withdraw", () => {
+  it("executes a deposit from public to confidential balance, transfer between confidential balances, and confidential to public balance", async () => {
+    const { sender: senderKeypair, recipient: recipientKeypair, spl } = await setup();
```
ananas-block: This test has no `expect` calls; a wrong balance passes as long as nothing throws. The Rust example asserts balance and UTXO count after each step. Assert the same here: `DEPOSIT_AMOUNT` after the deposit, `DEPOSIT_AMOUNT - TRANSFER_AMOUNT` after the transfer, `DEPOSIT_AMOUNT - TRANSFER_AMOUNT - WITHDRAW_AMOUNT` after the withdrawal.

sdk-libs/ts/src/client/prover/assembly.ts
```diff
@@ -59,31 +60,31 @@ export function circuitUtxo(value: object): CircuitUtxo {
   return result;
 }
 
-export function intoProver(
+export async function intoProver(
```
ananas-block: `intoProver` has no callers on main or on this branch. Delete it rather than porting it to async.

sdk-libs/ts/src/interface/codecs/index.ts
```diff
@@ -50,15 +50,15 @@ function writeDepositData(writer: Writer, value: DepositInstructionData): void {
...
+      writer.u8(1, "asset.kind").u8(asset.splInterfaceBump, "asset.splInterfaceBump");
     }
   }
   writer.u8(value.deposits.length, "deposits.length");
   for (const deposit of value.deposits) {
     writer
       .u8(deposit.assetIndex, "deposit.assetIndex")
       .bytes(deposit.viewTag, 32, "deposit.viewTag")
-      .bytes(deposit.owner, 32, "deposit.owner")
+      .bytes(deposit.recipientOwnerHash, 32, "deposit.owner")
```
ananas-block: The label still says `deposit.owner` after the field rename, so codec errors name a field that no longer exists. Use `"deposit.recipientOwnerHash"`.

sdk-libs/ts/src/client/prover/assembly.ts
tilo-14: wraps  Kit's getProgramDerivedAddress

ananas-block: the above comment describes that we should be able to remove getProgramDerivedAddress because it is redundant

ananas-block (dismissed):
Reviewed the full diff: endpoint resolution, the payerPublicKeyHash -> payer rename (semantically equivalent, verified against confidentialViewTag), the new confirm/poll path (no nested retries, error codes exist), and the ProofInputUtxo.fromKeypair destroy pattern (constructor clones the key). All sound.

Non-blocking follow-ups:
- `splInterfaceBump` in `sdk-libs/ts/src/interface/pda/index.ts` has no callers -- drop it or use it.
- Rust example transfer step uses `.utxos[0]` while the withdrawal step uses `.first()` -- prefer `.first()` per the no-indexing rule.
- The `createDeposit` viewTag switch to `confidentialViewTag()` is a real behavior fix; worth a unit assertion in wallet-actions.test.ts and a mention in the PR description.

ananas-block (approved):

PR #184, opened 2026-08-03
test: make test tiers structural, and replace the hygiene shell script

sergeytimoshin:
Three commits. The first two remove the reasons a test can silently not run; the third keeps a checker only for what the build system structurally cannot see.

`proofs = []` gated no code — `git grep 'feature = "proofs"' -- '*.rs'` returns nothing. It existed only to keep the proof-backed binaries out of `cargo test -p shielded-pool-tests`, and cargo skips a `required-features` target during `cargo check --all-targets`. So the isolation was bought by making those targets invisible to `check` and `clippy`, which is how the CU bench stopped compiling at #172 and went unnoticed for weeks.

A package boundary buys the same isolation without the blind spot:

| | isolated from fast tier | compiled by `check` | can be forgotten |
|---|---|---|---|
| `required-features` | yes | no | n/a |
| `#[ignore]` | yes | yes | yes, per test |
| separate package | yes | yes | no |


sergeytimoshin:
Superseded. Splitting this into its two halves showed they want opposite treatment.

**The package split is in #235.** `9daa72ce` (move the Groth16 suites into
`shielded-pool-proofs`) and `d0d74c4f` (the F-07 gate fix) were cherry-picked
verbatim -- git reports the suite moves as 100%/99% renames. #235 also documents
the hole they close: six targets, including `bench_cu`, carry
`required-features = ["proofs"]`, and CI's `just check-all` / `just clippy` pass
no `--features`, so cargo skips them under `--all-targets`. They compile on no
PR today.

**The checker half is being deleted rather than landed.** `6189082a`,
`6324273e` and `e45b0b46` replaced a 320-line shell script with 1243 lines of
xtask. The two checks it performs are better had from tools already in the repo:

- Orphaned test files: eliminated rather than policed. Flattening the suites to
  top-level `tests/*.rs` lets cargo auto-discover every one, so a file that
  never runs cannot be added -- exactly the property `shielded-pool-proofs`
  already relies on. That deletes `suite.rs` outright.
- Ledger citations: `cargo nextest list` is authoritative and proves a test
  runs, not merely that a function is defined somewhere. Run as a step in a job
  that already builds test binaries, it costs almost nothing, and it deletes
  `symbols.rs` and the symbol-index machinery.

`check_tallies` and `check_removal_claims` are dropped: a README count and a
retired entry's prose are review concerns, not CI gates.

The branch stays if any of this needs recovering.

PR #189, opened 2026-08-04
fix: slot-based indexer freshness gate, established once per sync

sergeytimoshin:
## Problem

A read-only CLI command `zolana utxos` over 3 utxos took **33.9 seconds** against the devnet
deployment. Photon answered individual queries in 250ms and proof generation took 3s,
so neither was the cost. Phase instrumentation put **33.8 of the 33.9 seconds** in the
indexer freshness gate and **13ms** in all the wallet crypto.

Two independent defects, both invisible without measurement.

### 1. The gate could not pass

`wait_for_indexer` compared the indexer's `Context.block_time` — a validator-assigned
block timestamp — against the caller's local wall clock. Block time trails real time,
so the first attempt routinely could not succeed *no matter how current the indexer
was*, and every gated request paid 400ms-2.5s of backoff. A correctness check had
degenerated into pure latency.

### 2. It fired per request, not per operation

The gate wrapped each individual indexer method, so one sync paid it **~36 times** —
per tag chunk, per fetch phase, per round.

## Root cause is the config type

```rust
pub struct IndexerRpcConfig {
    pub wait_for_indexer: bool,      // "wait"
    pub target_slot: Option<u64>,    // "...for what?"
}
impl IndexerRpcConfig { pub fn wait() -> Self { /* true, None */ } }
```

`Option<IndexerRpcConfig>` already expressed "no gate", so the bool was redundant —
and its redundancy is what encoded the unsound state (*wait, but for nothing in
particular*). That state was not merely representable, it was what the convenience
constructor produced. Collapsing both fields into one `require_slot: Option<u64>`
makes it unrepresentable: a freshness requirement **is** a slot.

## Changes

- **Slot comparison, no fallback.** Photon reports `max(slot)` alongside
  `max(block_time)` from the same scan of `blocks`, so it costs nothing. An indexer
  that does not report one is `IndexerSlotUnavailable`, not a silent downgrade.
- **Gate established once per sync** in both the blocking and async variants of
  `sync_wallet_with_config`.
- **CLI drops the pre-read gate entirely.** Every write command already waits for its
  own write via `confirm_private_transaction_sync` → `wait_for_indexed_transaction(signature)`,
  and merge waits for its leaf. Those waits are precise, so a blanket "is the indexer
  current?" check ahead of a read guarantees nothing extra. Three sync entry points
  collapse to one.
- **`SyncWalletConfig::new()` → `at_slot(slot)`** (same footgun: gate on, no target).

## Callers the type checker found

Deleting `IndexerRpcConfig::wait()` surfaced seven call sites, all wrong in ways
nobody would notice:

| site | problem |
| --- | --- |
| `discovery.rs`, `poll.rs` | gated **per page** inside cursor-pagination loops; `poll.rs`'s caller already retries against a deadline |
| `rfq.rs`, `deposit_transfer_withdraw.rs` (6) | passed the gate to `prove_transact`, which **skips** `get_merkle_proofs`' completeness poll — the one that waits until every requested leaf is present. They traded a strong guarantee for a weak, slow one |

All ungated.

## Measured against the devnet deployment

| | before | after |
| --- | --- | --- |
| `utxos` | 33.9s | **2.29s** |
| `balance` | ~28s | **2.31s** |
| `sync` | ~30s | **2.36s** |

The gated call inside a sync went from 4878ms to 417ms.

### DIFF
--- a/sdk-libs/client/src/retry.rs
+++ b/sdk-libs/client/src/retry.rs
@@ -68,6 +68,13 @@ impl IndexerPollConfig {
 pub struct IndexerRpcConfig {
     pub wait_for_indexer: bool,
     pub poll: IndexerPollConfig,
+    /// Chain slot the indexer must have reached before its answer is accepted.
+    ///
+    /// Read it from the same RPC the caller submits through, so both sides of the
+    /// comparison are slots on one chain. `None` falls back to the legacy
+    /// block-time comparison, which is unsound (see `Context::slot`) and only
+    /// remains for callers that cannot supply a slot.
+    pub target_slot: Option<u64>,
 }
 
 impl IndexerRpcConfig {
@@ -75,6 +82,16 @@ impl IndexerRpcConfig {
         Self {
             wait_for_indexer: true,
             poll: IndexerPollConfig::default(),
+            target_slot: None,
+        }
+    }
+
+    /// Wait until the indexer has persisted `slot`.
+    pub fn wait_for_slot(slot: u64) -> Self {
+        Self {
+            wait_for_indexer: true,
+            poll: IndexerPollConfig::default(),
+            target_slot: Some(slot),
         }
     }
 }
--- a/sdk-libs/client/src/retry.rs
+++ b/sdk-libs/client/src/retry.rs
@@ -68,21 +68,7 @@ impl IndexerPollConfig {
 pub struct IndexerRpcConfig {
     pub poll: IndexerPollConfig,
     /// Slot the indexer must have persisted before its answer is accepted.
-    ///
-    /// `None` accepts whatever the indexer currently has.
-    ///
-    /// There is deliberately no "wait, but for nothing in particular" state: a
-    /// freshness requirement IS a slot. The previous shape paired a
-    /// `wait_for_indexer: bool` with an optional target, which made the unsound
-    /// combination (wait with no target) not only representable but the value the
-    /// convenience constructor produced — it fell back to comparing the indexer's
-    /// block timestamp against local wall clock, mixing clock domains so the first
-    /// attempt could never pass while block time trailed real time. Every caller
-    /// that reached for it paid 400ms-2.5s of backoff per request for a check that
-    /// could not succeed.
-    ///
-    /// Read the slot from the same RPC the caller submits through, so both sides of
-    /// the comparison are slots on one chain.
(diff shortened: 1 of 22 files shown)


sdk-libs/client/src/retry.rs
```diff
@@ -66,15 +66,18 @@ impl IndexerPollConfig {
...
     pub poll: IndexerPollConfig,
+    /// Slot the indexer must have persisted before its answer is accepted.
+    /// `None` accepts whatever the indexer currently has.    
+    pub require_slot: Option<u64>,
 }
 
 impl IndexerRpcConfig {
-    pub fn wait() -> Self {
+    /// Require the indexer to have persisted `slot` before answering.
+    pub fn at_slot(slot: u64) -> Self {
```
ananas-block, line 77: wouldn't it easier to use timestamp instead of slot?
(we dont need to do an rpc call to get the current timestamp)

sdk-libs/indexer-api/src/lib.rs
```diff
@@ -492,6 +492,9 @@ impl<'de> Deserialize<'de> for Limit {
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct Context {
     pub block_time: i64,
+    /// Highest slot the indexer has persisted, for slot-based freshness checks.
+    #[serde(default)]
+    pub slot: Option<u64>,
```
ananas-block: makes sense, I think solana context is similar

ananas-block: is slot an option for solana context?

sdk-libs/client/src/retry.rs
sergeytimoshin: No, we shouldn't use timestamp - the problem is granularity: [`block_time`](https://solana.com/docs/rpc/http/getblocktime) measures in whole second but we have 2-3 slots / sec.

sdk-libs/indexer-api/src/lib.rs
sergeytimoshin: In Solana's [RpcResponseContext](https://docs.rs/solana-rpc-client-types/4.1.2/src/solana_rpc_client_types/response.rs.html#71) slot is non-optional. Agreed, make sense, slot is non-optional now.

ananas-block (approved):

PR #191, opened 2026-08-05
forester: expose Prometheus metrics and resolve config at the boundary

sergeytimoshin:
The forester runs unobserved. Liveness, whether the queue is draining, and whether it can still pay fees are answerable today only by reading CloudWatch logs. The fee-payer balance is the one that fails silently in production: the forester simply stops being able to submit.

`GET /metrics` on `--metrics-address`, Prometheus text format:

| metric | type | purpose |
| --- | --- | --- |
| `forester_last_run_timestamp` | gauge | liveness; the alert is staleness |
| `forester_sol_balance` | gauge | fee capacity, before it reaches zero |
| `queue_length` / `queue_capacity` | gauge | fill of the pending batch |
| `forester_transactions_failed_total` | counter | bucketed by reason |
| `forester_batches_submitted_total` | counter | work completed, per tree |
| `forester_indexer_response_time_seconds` | gauge | photon latency |
| `forester_indexer_proof_count` | gauge | ready zkp-batches |

Liveness and failures alone let a forester that loops without ever submitting look healthy, so `forester_batches_submitted_total` reports work completed. The contract's `forester_transactions_processed_total` is `epoch`-keyed, so rather than serve it with the label dropped, this adds a tree-keyed counter matching the dimensions `queue_length` already uses. Extending the contract is additive; the reasoning is recorded in its new `notes` field.

No new dependency. The forester is a small synchronous binary that pulls neither tokio nor hyper — its clients use `reqwest::blocking` and panic inside an async runtime. A `std::net::TcpListener` on a background thread avoids changing the binary's runtime model in order to expose seven gauges.

Every metrics failure is logged at debug and skipped rather than propagated, and a bind failure leaves the forester running without an endpoint. Reporting cannot break draining.

Six contract metrics are deliberately not served: `forester_epoch_detected`, `forester_epoch_registered`, `registered_foresters`, and three `epoch`-labelled transaction metrics. They describe the upstream Light Protocol epoch/registry design; this forester submits through a smart-account vault and has no epoch concept, so there is nothing to report. A gauge that is permanently zero reads as healthy data, which is worse than an absent one.

The contract file is left intact, and `contract_covers_served_names` asserts a subset rather than equality, because the k8s monitoring repo validates against the same file. Nothing in this crate validated against that file before, which is how six entries came to describe a design the forester does not have.

`RPC_URL`, `PHOTON_URL`, `PROVER_URL` and `PAYER` were read with `env::var` from inside `run` and `info`. Three costs: a missing variable surfaced partway through an operation, sometimes after network calls had been made; the same variable was read with different strictness in different places, with `PROVER_URL` required by `run` and optional in `info`; and neither entry point could be exercised without mutating process environment, so neither had unit tests.

`ForesterConfig::from_env()` resolves them once in `main`. `PAYER` is held unparsed, so a caller that never signs does not materialise key bytes. `start`, which does no work, no longer demands endpoints it never uses.

`forester.yml` ran `cargo check` only, so unit tests were compiled and never executed. It now runs `cargo test -p forester`.

Verification: `cargo fmt --all -- --check`, `./tools/check-test-hygiene.sh`, `cargo-machete`, `RUSTFLAGS="-D warnings" cargo check -p forester --all-targets` and `cargo clippy -p forester --all-targets` all clean; `cargo test -p forester` 20 passed, previously 8. The endpoint was exercised over a real socket: valid exposition text, 404 on unknown paths, counter accumulating per reason, balance reported in SOL rather than lamports. `forester start` with no environment succeeds; `forester info` with none fails with `RPC_URL is not set`.

Two follow-ups left out deliberately: a signer abstraction, since `PAYER` as a JSON byte array in the environment is a mainnet blocker and wants a remote or KMS signer; and decomposing `run()` at 695 lines, which is not worth doing while also changing behaviour.

### DIFF
--- a/.github/workflows/forester.yml
+++ b/.github/workflows/forester.yml
@@ -34,7 +34,7 @@ env:
 # integration tests will be reintroduced against the new combined tree type.
 jobs:
   unit:
-    name: forester / compile
+    name: forester / compile + test
     runs-on: ubuntu-latest
     if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
     steps:
@@ -43,3 +43,7 @@ jobs:
         with:
           private-libs-token: ${{ secrets.PRIVATE_LIBS_TOKEN }}
       - run: cargo check -p forester --all-targets
+        # The crate had no unit tests when this job was written, so it only
+        # compiled them. It now has coverage for config resolution and the
+        # metrics exposition format, which is worth nothing unless CI runs it.
+      - run: cargo test -p forester
--- a/forester/Dockerfile
+++ b/forester/Dockerfile
@@ -9,12 +9,23 @@
 # A bare `docker build forester/` will NOT work; the root .dockerignore keeps the
 # context small.
 #
-# rust:1.95: the workspace pulls solana crates that need a recent toolchain.
+# No build secret is needed. The workspace's three git dependencies
+# (Lightprotocol/groth16-solana, light-program-profiler, token) are all public,
+# so the PRIVATE_LIBS_TOKEN that CI's setup-rust action configures is not
+# required here. Verify with `git ls-remote` before adding secret plumbing.
 #
-# Pinned by digest, and to the same digests services/photon/Dockerfile uses: this
-# image is attested now, and a mutable tag lets one commit produce different
-# bytes depending on when it was built, which is exactly what the attestation is
-# supposed to rule out. Sharing photon's pins also shares the base layers.
+# Ships NO keys. The forester's signing keypair arrives at runtime as PAYER,
+# injected from Secrets Manager by the ECS task definition.
+
+# Pinned by digest, to the same digests services/photon/Dockerfile uses: a
+# mutable tag lets one commit produce different bytes depending on when it was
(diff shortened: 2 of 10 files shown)


sergeytimoshin:
Rebased onto `main` (`31aa6f40`); `MERGEABLE`, so CI runs against current main for the first time since 12 Aug — the previous green referred to a base that no longer exists.

One conflict, in `forester/Dockerfile`. This branch replaced the builder base with an unpinned `rust:1.97-slim-bookworm`; `main` has since pinned it by digest to the same image `services/photon/Dockerfile` uses, so one commit cannot produce different bytes depending on when it was built. Resolved by keeping the digest pin and this branch's non-root user and `EXPOSE 9998`, which `main` still lacks — the runtime image runs as root today.

The 1.97 comment is preserved in substance: `rust-toolchain.toml` pins 1.97.0, so rustup downloads it over the 1.95 base on every build. ~30s, and moving the FROM to a 1.97 digest would unshare photon's layers.

`cargo test -p forester` 21 passed, clippy and rustfmt clean.

ananas-block (approved):

PR #192, opened 2026-08-05
feat(cli): load release accounts with local programs

tilo-14:
## Summary
- Add `zolana dev start --release-accounts`.
- Load pinned release account snapshots while retaining local programs and services.
- Preserve existing release and fully local behavior.

## Test plan
- [x] `cargo test -p zolana-cli release_selection_separates_accounts_from_programs`


sergeytimoshin:
Wdyt about to reuse light-protocol's approach and load all the needed "release" accounts by default without unnecessary flags?

PR #193, opened 2026-08-05
chore: review and cleanup keyderivation

ananas-block:
**Motivation:**
1. ensure ecosystem compatibility of key derivation

**Changes:**
1. add `from_keypair`  replaces `from_solana_keypair` and `from_ed25519`
    - derive viewing key and nullifier key from signature for ED25519 and from key exchange for P256
2. add ShieldedPda
3. rename `SignatureType` -> `Curve`
4. consolidate derivation into derivation.rs
5. move `key_schedule` to encryption.rs, move remaining merge helpers to zolana transaction
6. implement solana Signer trait for ShieldedKeypair 
7. update spec
8. update comments
9. add HSM compatibility test

**Notes:**
1. hardware wallet clients should implement the ShieldedKeypair trait
2. seed based wallets do their own key management and should implement the ShieldedKeypair trait


sdk-libs/keypair/src/shielded.rs
```diff
@@ -69,67 +73,66 @@ impl AsRef<NullifierKey> for ShieldedKeypair {
...
-            signing_key,
-            nullifier_key,
-            viewing_key,
-        })
+impl Signer for ShieldedKeypair {
+    fn try_pubkey(&self) -> Result<solana_address::Address, SignerError> {
+        let bytes = self
+            .signing_pubkey()
+            .as_ed25519()
+            .map_err(|e| SignerError::Custom(e.to_string()))?;
```
ananas-block: try to use proper error conversion

sdk-libs/keypair/src/shielded.rs
```diff
@@ -69,67 +73,66 @@ impl AsRef<NullifierKey> for ShieldedKeypair {
...
+        Ok(solana_keypair::Signature::from(sig))
     }
 
+    fn is_interactive(&self) -> bool {
+        false
+    }
+}
+
+impl ShieldedKeypair {
     pub fn new() -> Result<Self, KeypairError> {
```
ananas-block: add new_p256, new eddsa

sdk-libs/keypair/src/signing_key.rs
```diff
@@ -40,13 +41,13 @@ impl Clone for SigningKey {
...
         Self {
             inner: SigningKeyInner::P256(SecretKey::random(&mut OsRng)),
         }
     }
 
-    /// A fresh random ed25519 key. Mirrors [`Self::new`] (P256) for callers that
+    /// A fresh random ed25519 key. Mirrors [`Self::new_p256`] for callers that
     /// need a throwaway key on the ed25519 rail; the secret bytes are zeroized
     /// once copied into the dalek key.
     pub fn new_ed25519() -> Self {
```
ananas-block, line 39: rename to new

sergeytimoshin:
1. Spec vs. implementation mismatch on nullifier HKDF info strings (medium). docs/spec.md (lines 321, 323) documents the canonical derivation as info="TSPP/nullifier/ecdh/v1" and "TSPP/nullifier/ed25519/v1", but the implementation in sdk-libs/keypair/src/derivation.rs uses TSPP/nf_key/ecdh/v1 and TSPP/nf_key/ed25519/v1. Because the client derivation is explicitly not proof-enforced (circuits constrain only nullifier_pk = Poseidon(secret)), correctness depends entirely on every implementation using the same tag. A second wallet built from the spec would derive a different nullifier_secret → different nullifier_pk → different owner_hash, breaking the "derive the same secret from any wallet" recovery guarantee. Either the spec or the constants should change so they agree. (Note the tests in tests/cases/nullifier.rs were written against the code strings, so they pass and don't catch the drift.)

2. Viewing-key derivation in the spec doesn't match the code path. spec.md line 341 says an Ed25519 owner's viewing key comes from ViewingKey::from_seed(wallet_seed, account) with wallet_seed := HKDF(derive_sig, "TSPP/viewing_seed/ed25519/v1"). The actual ShieldedKeypair::from_keypair path instead calls RoleExpansion::viewing_key(), which expands info="TSPP/view_key/ed25519/v1" directly from the seed to 48 bytes. The tag TSPP/viewing_seed/ed25519/v1 appears nowhere in the code, and from_seed isn't on this path at all. 

3. pda.rs references a design doc that doesn't exist. The module doc points to docs/shielded_pda_design.md; that file isn't in the tree (the docs dir has spec.md, properties.md, etc., but not this one). Either add it or fix the reference.

sergeytimoshin:
Second pass. Nothing has landed since `55e78bd` (2026-08-06), so the three items from the previous round are unchanged and repeated below; items 4-8 are new.

## Build state

Ran locally against `55e78bd`, since this branch has never had a CI run:

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p zolana-keypair` — 23 passed (11 `keypair`, 9 `hsm_backends`, 3 unit)
- `cargo test -p zolana-transaction` — all passed

The 113-file rename sweep compiles and the unit suites are green. `cargo test-sbf` not run for the `sdk-tests` programs.

## 1. Nullifier HKDF info strings differ between spec and code

`docs/spec.md` 321/323 gives `info="TSPP/nullifier/ecdh/v1"` and `info="TSPP/nullifier/ed25519/v1"`. `derivation.rs` uses `TSPP/nf_key/ecdh/v1` and `TSPP/nf_key/ed25519/v1`.

The client derivation is explicitly not proof-enforced — circuits constrain only `nullifier_pk = Poseidon(secret)` — so correctness rests entirely on every implementation agreeing on the tag. A second wallet built from the spec derives a different `nullifier_secret`, hence a different `nullifier_pk` and `owner_hash`, and the "derive the same secret from any wallet" recovery property breaks. `tests/cases/nullifier.rs` was written against the code strings, so it passes and does not catch the drift.

Same drift on the point name: the spec calls it `P_nullifier`, the code calls it `P_derive` (`DST_DERIVE_P_DERIVE`, `P_DERIVE_SEC1`, `p_derive()`), while the DST string itself says `P_nullifier`. Three names for one point.

## 2. Viewing-key derivation in the spec is not the code path

spec.md 341 says an Ed25519 owner's viewing key comes from `ViewingKey::from_seed(wallet_seed, account)` with `wallet_seed := HKDF(derive_sig, "TSPP/viewing_seed/ed25519/v1")`.

`ShieldedKeypair::from_keypair` instead goes through `RoleExpansion::viewing_key()`, which expands `info="TSPP/view_key/ed25519/v1"` from the seed straight to 48 bytes. `TSPP/viewing_seed/ed25519/v1` appears nowhere in the tree, and `from_seed` is not on this path.

`from_seed` now has no production caller at all — the only use left is `sdk-tests/zk-program-swap/sdk/src/state/order.rs:282`. Either the spec should describe `RoleExpansion`, or `from_seed` should be deleted; documenting a test-only constructor as *the* constructor is the worst of the two.

## 3. `pda.rs` points at a design doc that isn't in the tree

`sdk-libs/keypair/src/pda.rs:1` cites `docs/shielded_pda_design.md`. `docs/` has `spec.md`, `properties.md` and friends, not that file.

## 4. Only one of three committed points is pinned to its DST

`tests/cases/nullifier.rs:71` regenerates `P_derive` via `NistP256::hash_from_bytes::<ExpandMsgXmd<Sha256>>(&[b""], &[DST])` and asserts it equals `P_DERIVE_SEC1`. `P_CONST_SEC1` and `P_PDA_SEC1` are bare constants — their DSTs appear only in the pairwise-distinctness list, which would pass just as happily against an arbitrary point.

I checked all three; every one is correct:

```
P_derive  TSPP/nullifier/P_nullifier/v1  039ef16592429da1403eaa29058eb7d9d5ad15a2ea557174f7b01ff7fe484eeeaf
P_const   TSPP/view_root/P_const/v1      030e4df946bce14b95292f13e133d2b0c64e898b5644f620a5bed25a061a42fcdb
P_pda     TSPP/pda_root/P_pda/v1         038a31d53c5ad20d1cd7ee1fbb9927bd0cdfb61b1b89f6b2c5a94a5f081fe16b5b
```

So this is prophylactic rather than a bug. It is still worth closing, because the failure mode is silent and total: `P_pda` roots every single-holder PDA identity through `from_viewing_key`, and `P_const` roots every view-tag and transaction-viewing secret. A point with a known discrete log in either slot is indistinguishable from a correct one by every test currently in the suite. Generalising `p_derive_matches` into a loop over the three pairs is a few lines.

## 5. `ecdh(P_const)` is permitted, and `view_root` is exactly that value

`is_derivation_point` covers `P_derive` and `P_pda` but deliberately excludes `P_const`, on the stated grounds that "`view_root` expands per-purpose secrets, not identity roles". `tests/cases/nullifier.rs:180` pins this, naming `P_const` the `benign` counterparty and asserting `vk.ecdh(&benign).is_ok()`.

But `view_root = HKDF-Extract(∅, ECDH(viewing_sk, P_const))`, so `vk.ecdh(P_const)` returns precisely the IKM that root is extracted from. Whoever holds that 32-byte value computes `view_root`, and from it `sender_view_tag_secret`, `recipient_view_tag_secret` and `tx_viewing_secret` — the entire scanning and decryption half of the identity. Those per-purpose secrets are the privacy-critical material, so the justification for the exclusion argues the wrong way round.

Latent, not exploitable today: no caller outside the crate passes an attacker-supplied point to `ecdh` (the only external use is `sdk-libs/transaction/benches/wallet_ops.rs:85`). But `ecdh` is `pub`, it is on `ViewingKeyTrait`, and `ShieldedPda::from_key_exchange` already takes a counterparty pubkey off the wire.

The fix is free: `view_root` calls `ecdh_x_point` directly rather than going through `ecdh`, so adding `P_const` to `is_derivation_point` breaks nothing.

## 6. The derivation-input guard is vacuous on the P-256 `sign` arm

`SigningKey::sign` checks `is_derivation_input(msg)`, but on the `P256` arm `msg` is a prehash handed to `sign_prehash`. The check can only fire on a 32-byte digest whose leading bytes happen to spell `TSPP/derive/`.

Harmless — the P-256 rail's seed is `ECDH(sk, P_derive)`, not a signature, and that path is guarded by `is_derivation_point` — but the guard reads as protection it does not provide there. Worth a comment saying the check exists for the Ed25519 arm.

## 7. AWS SDK in a core crypto crate's test build

`aws-sdk-kms = "1"` and `aws-smithy-mocks = "0.3"` are the only dependencies in `sdk-libs/keypair/Cargo.toml` not taken from the workspace, so they float independently of the pinning everything else obeys. `cargo test -p zolana-keypair` now pulls the AWS SDK tree.

The HSM tests are worth having. Suggest workspace-pinning both, and considering whether they belong behind a feature or in a separate crate so the keypair test loop stays light.

## 8. `view_root` is computed eagerly (minor)

`ViewingKey::from_secret_key` always computes `view_root`, so every `ViewingKey` pays a scalar multiplication whether or not it ever derives a view tag. `get_transaction_viewing_key` builds a fresh single-use key per transaction and only uses `self.secret` for encrypt/decrypt — that ECDH is always wasted. A `OnceCell` would make it lazy. Low priority; noting it because the cost is on the transaction-building path.

sdk-libs/keypair/src/shielded.rs
ananas-block: Done in 2479bee3: added `impl From<KeypairError> for SignerError` in `error.rs`, so the three sites here use `?`/`.into()` instead of `SignerError::Custom(e.to_string())`. Also switched `try_pubkey` to report `NoSolanaAddress` for a P-256 owner; it was surfacing `InvalidSignatureType(0)` from `as_ed25519`, which reads as a malformed key rather than a key on the wrong rail.

sdk-libs/keypair/src/shielded.rs
ananas-block: Done in 2479bee3: `ShieldedKeypair::new()` is gone, replaced by `new_p256()` and `new_ed25519()`, and the 114 call sites across 24 files now name the rail. `new_ed25519` needed no new derivation code -- `from_keypair` already handles that rail through `derivation_seed`. Covered by `random_constructors_pick_their_rail`.

sdk-libs/keypair/src/signing_key.rs
ananas-block: Leaving `new_p256` / `new_ed25519` as they are. The note is anchored on `new_ed25519` but the change under it was `new()` -> `new_p256()`, and reverting that would reintroduce the implicit-rail default we just removed from `ShieldedKeypair`. Both types now name the rail in the constructor.

ananas-block:
@sergeytimoshin thanks -- addressed in 2479bee3. Rundown by item.

**1, 2, 3 -- already fixed, no action.** The branch moved after your second pass: `53049f8d` landed at 20:42 on the 13th. `docs/spec.md:484,492` now read `TSPP/nf_key/ed25519/v1` and `TSPP/nf_key/ecdh/v1`, matching `derivation.rs`; `spec.md:466-470` describes the `prk := HKDF-Extract(derivation_seed)` role expansion, so `TSPP/viewing_seed/ed25519/v1` is gone from the tree; and `docs/shielded_pda_design.md` exists. On the point name, `spec.md:491` spells it `P_derive` with `DST="TSPP/nullifier/P_nullifier/v1"`, exactly as the code does -- the DST literal cannot change without rotating the point, so that one stays.

Declined the `from_seed` deletion. It is not dead: the TS SDK implements the same `TSPP/seed/p256_viewing` derivation at `sdk-libs/ts/src/keypair/viewing-key.ts:116` and calls it from `shielded.ts:221`. Deleting the Rust side would break parity.

**4 -- fixed.** `p_derive_matches` is now `committed_points_match_dsts`, looping over all three `(DST, SEC1)` pairs and regenerating each, plus pairwise distinctness. All three constants were correct, so nothing moved.

**5 -- fixed.** `P_CONST_SEC1` is in `is_derivation_point` and the rationale comment is rewritten. Free as you said: `view_root` goes through `ecdh_x_point`, so no internal caller broke; only the `benign` counterparty in the guard test needed retargeting to a random point. Both `SigningKey::ecdh` and `ViewingKey::ecdh` now reject all three committed points.

**6 -- documented, not changed.** Correct that the check is inert on the P256 arm. Added a doc line saying so and pointing at `is_derivation_point` as the actual guard for that rail. No behavior change; a digest colliding with the prefix is a 2^-96 event.

**7 -- fixed, and there were five, not two.** `solana-derivation-path`, `solana-seed-phrase` and `ed25519-dalek-bip32` were floating too. All five are in the workspace table now, `Cargo.lock` unchanged. Skipped the feature gate: optional dev-dependencies are not expressible in Cargo, and `hsm_backends` is already its own `[[test]]` target, so `--test keypair` does not link the AWS tree.

**8 -- fixed.** `view_root` is a `OnceLock` resolved in `derive_secret32`. `get_transaction_viewing_key` returns a key used only for encrypt/decrypt, so it now skips the scalar multiplication entirely.

Two things beyond your list, both from the author threads: `impl From<KeypairError> for SignerError` replaces the three stringify sites, and `try_pubkey` now reports `NoSolanaAddress` for a P-256 owner instead of `InvalidSignatureType(0)`. `ShieldedKeypair::new()` is gone in favour of `new_p256()` / `new_ed25519()`.

Separately, and out of scope here: the TS SDK derives a different identity than Rust for the same ed25519 wallet. `sdk-libs/ts/src/keypair/shielded.ts:217` uses `HKDF(signing_secret, "TSPP/nullifier")` plus `ViewingKey.fromSeed`, where Rust uses the derivation-seed role expansion -- different `nullifier_pk`, different `owner_hash`. That is the live version of the drift item 1 warned about. Not touched by this PR.

Verified: `cargo test -p zolana-keypair` / `-p zolana-transaction` / `-p zolana-wallet` / `-p zolana-cli` / `-p shielded-pool-tests`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo doc` (which also fixed a pre-existing broken `PublicKey::hash` intra-doc link).

sergeytimoshin (dismissed):

sergeytimoshin (approved):

PR #194, opened 2026-08-05
docs: publish versioned TypeScript API reference

tilo-14:
## Summary
- generate strict TypeDoc reference pages for every public `@heliuslabs/zolana` entry point
- rebase onto `main` after the `@heliuslabs/zolana` rename
- rewrite the npm README (Solana Rings, localnet client, private transfer) and set the package version to `0.1.0-alpha`
- validate documentation in TypeScript CI and publish immutable versions from `ts-sdk-v*` tags (prerelease may be latest until a stable exists)
- document the current GitHub Pages URLs and planned `www.helius.dev/privacy/api` migration

## Test plan
- [x] `npm run docs:ts`
- [ ] `npm run check:ts`
- [ ] TypeScript CI on this PR


sdk-libs/ts/README.md
```diff
@@ -106,43 +100,56 @@ console.log(wallet.balance(SOL_MINT).amount);
...
+the same owner seed, as shown above.
+
+### Endpoints
+
+**Devnet:**
+
+| Service    | Host the SDK uses                                                   | Notes                                                                                     |
+| ---------- | ------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
+| Solana RPC | `https://devnet.helius-rpc.com/?api-key=<API_KEY>`                  | Helius key. Fund the payer with [devnet SOL](https://www.helius.dev/docs/rpc/devnet-sol). |
+| Indexer    | `http://zolnet-devnet-1779374825.eu-north-1.elb.amazonaws.com`      | Fetches encrypted state.                                                                  |
```
sergeytimoshin, line 124: I would prefer to not merge it this way and wait for gatekeeper (should be unblocked today).

sergeytimoshin (dismissed):

Atamanov (changes requested):
**Request changes.** Publishes the versioned TypeScript API reference and rewrites the npm README. Four items block merge, each verified against the SDK source and the npm registry.

- Docs version collides with a published npm artifact. `sdk-libs/ts/package.json` sets `0.1.0-alpha` and the workflow gates on tag==version, so docs publish under `0.1.0-alpha`. That npm version's `exports` lack `./ring`; the ring surface ships in `0.1.2-alpha` (the `alpha` dist-tag). The immutable page would document `./ring` the matching artifact does not expose. Bump to a new unpublished version and generate docs from that release commit.
- README install resolves to the wrong artifact. `npm install @heliuslabs/zolana` (README:18) takes `latest`, which is `0.1.0-alpha` without `./ring`. Document `@heliuslabs/zolana@alpha` until `latest` advances.
- Wallet creation does not register on chain (README:168-170). `new Wallet()` builds local state only; registration is a separate signed `buildRegistrationTransaction` the quick start never sends, so a recipient that only constructed a wallet fails a transfer with `WALLET_RECIPIENT_NOT_REGISTERED`. State registration as an explicit transaction.
- The SDK does not start localnet (README:136-138). It spawns no processes; `createZolanaClient({})` resolves localhost endpoints only. Require the caller to start the stack first, e.g. `zolana dev start`.
- Endpoints section contradicts itself (README:112 vs 128-133): "needs one URL" then requires three, and the devnet example sets `allowInsecureHttp: true` over plaintext `http://` while the notes require HTTPS. Reconcile, and scope the insecure path to devnet test funds.

Hold stands with the maintainer request to wait for the gatekeeper on the README thread, consistent with the version issue above.

Atamanov (dismissed):

Atamanov (approved):

PR #198, opened 2026-08-07
sdk: resume each tag where the last sync left it, and report a cursor on short pages

sergeytimoshin:
Two halves of one mechanism: the client remembers where each tag was read to, and photon reports a position even on the last page. Separately, neither is complete — a client that syncs to the tip cannot record where the tip was.

## Client: per-tag cursors

A tag discovered late was skipped permanently. Sync kept a single cursor shared by every view tag. Tags are discovered lazily, as transactions decrypt, so a tag learned in a later round was queried with a cursor already advanced past its transactions, and those transactions were never returned again. The motivating case is one account on two devices: device B derives tags from its own counter, outside device A's initial window, so A advances the shared cursor past B's history. That is a permanently wrong balance, not a slow sync.

Every round also refetched the whole history. Rounds exist so that tags found by decrypting round N can be queried in round N+1, but each round re-queried every tag from position zero. On devnet, round 1 refetched all 222 transactions to find a handful of new tags.

Cursors are now keyed by tag. Tags sharing a resume point are grouped into one request, so the request count is unchanged in the common case where every tag sits at the same position. Tags read to the tip are recorded for the duration of a sync and skipped in later rounds; that set is not persisted, since "the tip" means the tip as of now. Transactions and encrypted utxos keep separate sets, because reaching the tip of one stream says nothing about the other.

## Server: report the last row's position on a short page

`next_cursor_from_rows` returned `None` when a page came back shorter than the limit. It now always reports the last row's position.

```rust
rows.last().map(cursor_from_row).transpose().map(|c| c.map(Base64String))
```

The cursor does double duty. As a page token, "short page means no more pages" is correct and `None` is a fine answer. As a sync watermark it is the client's record of how far it has read, and dropping it on the last page is exactly when the client most needs it.

No schema change, deliberately. Response types are `#[serde(deny_unknown_fields)]`, so adding a field would break every deployed client. This changes what an existing optional field contains, not its shape: `next_cursor` was already `Option`, callers already handled both cases, and a cursor pointing at the last row of a short page is a valid cursor — paging from it returns an empty page, which is the correct terminal state.

Callers can also stop one round trip earlier, because a short page is already known to be the last:

```rust
cursor = response.next_cursor;
if last_page || cursor.is_none() { break; }
```

## Soundness

Persisting a watermark across syncs is sound only if photon cannot insert a row behind a position it has already returned. It cannot: block batches commit one at a time, in slot order, inside a single transaction (`index_block_batch`), and there is no re-index path. Photon indexes at `Confirmed` and never re-indexes a rolled-back slot; that exposure predates this change.

## The mock was hiding both halves

The test mock returned a cursor for every non-empty page, while photon returned `None` for any page shorter than the limit (`next_cursor_from_rows`). So the mock let a client look incremental in tests while rescanning from zero in production — `tag_cursors=0` after a round that fetched 222 rows. Making the mock faithful to the server broke the per-tag cursor test on the first run, which is what surfaced the server half.

`a_short_page_still_reports_its_last_row` and `an_empty_page_ends_the_stream` pin both directions: a short page must carry a cursor, an empty page must not.

## Measured

Devnet, wallet of 222 transactions. Balance verified identical against a build of `main` (99600000 both), and a transfer settled exactly (-1000000 sender, +1000000 recipient).

| | before | after |
|---|---|---|
| sync total | 3427ms | ~2700ms |
| round 1, tags | 749ms | ~205ms |
| round 1, proofless | 662ms | ~124ms |

Sync cost no longer grows with history.

Nullifier queries still restart from zero every round and are now the largest remaining repeat, ~510ms of round 1. The same treatment applies; left out to keep this reviewable.

### DIFF
--- a/sdk-libs/transaction/src/wallet/state.rs
+++ b/sdk-libs/transaction/src/wallet/state.rs
@@ -145,18 +145,7 @@ pub struct Wallet {
     pub nullifiers: HashSet<[u8; 32]>,
     pub last_synced: i64,
     /// Per-view-tag sync watermarks: for each tag, the indexer cursor up to which
-    /// every matching transaction has already been seen.
-    ///
-    /// Per tag rather than one shared position, because the tag set GROWS. Tags
-    /// are derived from a local counter plus a window, and a second device
-    /// spending beyond that window makes the counter lag by more than the window
-    /// absorbs. A tag learned late must be scanned from the beginning even though
-    /// other tags have advanced far past the slots involved; a single cursor would
-    /// skip those transactions permanently.
-    ///
-    /// Kept in memory only. A client that persists wallet state across restarts
-    /// should persist this with it -- without it, sync is correct but pays for the
-    /// full history every time.
+    /// every matching transaction has already been seen.    
     pub sync_cursors: HashMap<ViewTag, Vec<u8>>,
 }
 
--- a/sdk-libs/transaction/src/wallet/state.rs
+++ b/sdk-libs/transaction/src/wallet/state.rs
@@ -1,7 +1,7 @@
 use std::collections::{hash_map::Entry, BTreeSet, HashMap, HashSet};
 
 use solana_address::Address;
-use zolana_keypair::{shielded::ShieldedAddress, P256Pubkey};
+use zolana_keypair::{shielded::ShieldedAddress, viewing_key::ViewTag, P256Pubkey};
 
 use crate::{
     error::TransactionError, instructions::transact::OutputContext, utxo::Utxo, AssetRegistry,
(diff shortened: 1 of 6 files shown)


sergeytimoshin:
## Stress test on devnet

Ran the `xtask loadtest` generator against the deployed devnet with this branch merged in. Same generator, same 4 workers, same 200000 lamports/transfer as the earlier baseline. The baseline ran 300s, this ran 600s -- and against a devnet whose history had grown in between, so the sync phase had *more* to do, not less.

| | baseline | with this change |
|---|---|---|
| sync p50 | 5482ms | 3422ms |
| sync p95 | 7941ms | 4432ms |
| sync mean | 5785ms | **3432ms** |
| prove mean | 4472ms | 4196ms |
| send mean | 1165ms | 1097ms |
| confirm mean | 360ms | 398ms |
| total mean | 11783ms | **9123ms** |
| throughput | 0.27 tps | **0.39 tps** |
| ok / failed | 91 / 0 | 235 / 0 |

Sync is 41% faster and end-to-end transfer latency is 23% faster. Sync was the largest phase before this change and is now smaller than prove.

Zero application failures across 235 transfers. The 24 "throttled" are Helius returning nginx 403s upstream, unrelated to this change -- the generator backs off and retries.

Correctness checks alongside the load: balances verified identical to a build of `main` (99600000), and a real transfer settled exactly (sender -1000000, recipient +1000000, recipient having already synced before the transfer -- so the incremental path did pick up new data).

ananas-block (approved):

PR #201, opened 2026-08-07
client: fetch spend and dummy nullifier proofs concurrently

sergeytimoshin:
Two independent indexer round trips that ran back to back: different methods, neither consuming the other's output. 317ms and 114ms on devnet, so the pair cost 431ms where it need only cost 317ms.

Needs `R: Sync` on the submission path since the scoped threads share the indexer, widened through the two signing entry points that call it.

Same caveat as #200: a single-client latency win, not a throughput win.


sergeytimoshin:
Folded into #200. Same change, same evidence — reviewing it separately meant re-reading the other half's justification. No work is lost: the branch it targeted now contains these commits.

PR #204, opened 2026-08-08
prover: publish how long a job waits to be picked up

sergeytimoshin:
Stacked on `prover/metrics-gauges` (#196).

A client-observed prove phase of 118s was reconciled against 0.28s of compute and a worker never seen running more than one job. Nothing measured the time between a job being accepted and being picked up, so there was no way to separate queue wait from the client polling too slowly for its own result.

`prover_queue_wait_seconds_{last,mean,max}`, by queue, stamped from the job's `CreatedAt` at dequeue — the same field the expiry check already relies on. Recorded before the expiry branch, so expired jobs, which are the extreme case, still count.

`prover_queue_wait_time_seconds` already exists as a histogram and stays. The CloudWatch agent drops histograms entirely — not as `_bucket`, `_sum` or `_count` — so nothing derived from it reaches a dashboard. After a run that generated 125 proofs, the `zolnet/prover` namespace contained every gauge and counter and no histogram series. These gauges carry the same information in a shape that survives the trip.

Deployed to devnet, the metric read 105,080ms mean against a client-observed prove phase of 104,688ms. That near-exact agreement established that essentially all of the prove time was queue wait rather than compute or client polling, and pointed the investigation at dispatch, where an O(n) scan was capping admission at 0.88 jobs/s (#202).

Tests assert on the rolling window rather than the gauges: a Prometheus Gauge is write-only, and reading one back would mean a dependency on `testutil` for no extra coverage, since the window is where mean and max are computed. They cover the mean/max arithmetic, that queues do not share a window, and that a backwards clock cannot publish a negative wait — which would read as a queue returning results before they were submitted.

### DIFF
--- a/prover/server/server/metrics.go
+++ b/prover/server/server/metrics.go
@@ -146,6 +146,37 @@ var (
 		[]string{"circuit_type"},
 	)
 
+	// How long a job sat in Redis between being accepted and being picked up.
+	//
+	// The gap nobody measured. A client-observed "prove" phase of 118s was
+	// reconciled against 0.28s of compute and an idle worker, and there was no
+	// way to tell whether the missing two minutes were queue wait or the client
+	// polling too slowly for its result. Stamped from the job's CreatedAt at
+	// dequeue, which the expiry check already relies on.
+	QueueWaitLast = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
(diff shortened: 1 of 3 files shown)


sergeytimoshin:
Folded into #196. Same change, same evidence — reviewing it separately meant re-reading the other half's justification. No work is lost: the branch it targeted now contains these commits.

PR #205, opened 2026-08-08
indexer: report the last row's position even on a short page

sergeytimoshin:
Stacked on `sdk/incremental-sync` (#198).

`next_cursor_from_rows` returned `None` when a page came back shorter than the limit. It now always reports the last row's position.

```rust
rows.last().map(cursor_from_row).transpose().map(|c| c.map(Base64String))
```

The cursor does double duty. As a page token, "short page means no more pages" is correct and `None` is a fine answer. As a sync watermark it is the client's record of how far it has read, and dropping it on the last page is exactly when the client most needs it. The effect was that a client syncing to the tip could not remember where the tip was: next sync it re-read from its previous position, and the cost of a sync grew with total history rather than with what had arrived since.

No schema change, deliberately. Response types are `#[serde(deny_unknown_fields)]`, so adding a field would break every deployed client. This changes what an existing optional field contains, not its shape: `next_cursor` was already `Option`, callers already handled both cases, and a cursor pointing at the last row of a short page is a valid cursor — paging from it returns an empty page, which is the correct terminal state.

Callers can stop one round trip earlier, because a short page is already known to be the last:

```rust
cursor = response.next_cursor;
if last_page || cursor.is_none() { break; }
```

The cursor is kept for the next sync; only the extra end-of-stream request goes away.

`a_short_page_still_reports_its_last_row` and `an_empty_page_ends_the_stream` pin both halves: a short page must carry a cursor, an empty page must not.

How the gap survived: the client-side mock always returned a cursor, while photon returned `None` on short pages. Making the mock faithful to the server made the guard test fail immediately.

### DIFF
--- a/sdk-libs/wallet/src/wallet_sync.rs
+++ b/sdk-libs/wallet/src/wallet_sync.rs
@@ -955,18 +955,12 @@ mod tests {
                 .collect();
             rows.sort_by_key(|(slot, _)| *slot);
 
-            // Photon's rule (`next_cursor_from_rows`): a page shorter than the
-            // limit is the end of the stream and carries no cursor. Modelling
-            // this matters -- a mock that always returns a cursor lets a client
-            // look incremental in tests while rescanning from zero in
-            // production.
-            let short_page = rows.len() < limit;
+            // Photon's rule (`next_cursor_from_rows`): the position of the last
+            // row returned, or None when there were none. A short page still
+            // carries a cursor, so the client can resume from the tip next sync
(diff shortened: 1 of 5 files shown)


sergeytimoshin:
Folded into #198. The client cursors and the short-page cursor are two halves of one mechanism — a client that syncs to the tip cannot record where the tip was without both — so reviewing them apart means reviewing half a change. `sdk/incremental-sync` now contains these commits; no work is lost.

PR #206, opened 2026-08-08
xtask: report wallet depth and emit a machine-readable summary

sergeytimoshin:
Stacked on `xtask/loadtest` (#197).

Two additions to the load generator, both to stop it misreporting what it measured.

`warmup_sync_ms` is a wallet-depth proxy. A wallet's sync cost scales with the history it already has, and every run leaves hundreds more transfers behind on the same wallets, so runs taken hours apart are not comparable — and nothing in the output said so.

That is not hypothetical. An earlier conclusion that throughput scales near-linearly with workers came from comparing an 8-worker and a 16-worker run taken hours apart. Re-running 8 workers immediately after gave 0.67 tps against the earlier 1.04, and the difference was accumulated wallet history, not concurrency. The report now prints each worker's first sync and warns in place:

```
wallet depth:
  first sync   mean     3396ms   across 8 wallets
  Sync cost scales with a wallet's history, and every run leaves more
  behind on these wallets. Compare runs only at similar first-sync cost.
```

`--json` writes phase percentiles, counts and throughput to a file, so a series of runs can be diffed rather than eyeballed out of scrollback. Every before/after figure in the current performance work came from these files.

The generator is a measuring instrument, and one that silently drifts between readings is worse than none. Both changes make a misleading comparison visible at the point of comparison rather than in hindsight.

### DIFF
--- a/xtask/src/loadtest.rs
+++ b/xtask/src/loadtest.rs
@@ -63,6 +63,13 @@ struct Stats {
     /// failed" says nothing about whether the prover is down, the wallet is
     /// empty, or the indexer is behind.
     errors: std::collections::BTreeMap<String, u64>,
+    /// This worker's first sync, before it sent anything.
+    ///
+    /// Stands in for how much history its wallet already carried. Sync cost
+    /// scales with that, and every run leaves hundreds more transfers behind on
+    /// the same wallets, so throughput is only comparable between runs at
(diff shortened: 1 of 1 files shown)


sergeytimoshin:
Folded into #197. Same change, same evidence — reviewing it separately meant re-reading the other half's justification. No work is lost: the branch it targeted now contains these commits.

PR #207, opened 2026-08-09
sdk: resume the proofless scan, type indexer failures, and stop unbounded proof polling

sergeytimoshin:
Four independent changes, one per commit. Based on `stress/watermark`, the tree the load generator runs from; needs rebasing onto a reviewable base before merge.

## 1. Remember where the proofless scan stopped

`sync_wallet` keeps per-tag cursors for the transaction stream, so each sync resumes where the last finished. The encrypted-utxo stream, where proofless deposits come from, had no such memory: its `scanned_to_tip` set was a local, created fresh on every call, so every sync paged the entire history for every tag and discarded all but the deposits client-side.

Devnet, `ZOLANA_TIMING`, 8 workers, wallets carrying ~10k transfers:

```
sync_wallet_total             2650ms
  fetch_round                  970ms
  fetch_proofless_deposits     909ms   (to keep 2.5 rows)
  collect_sort_and_decrypt     341ms
  fetch_shielded_by_nullifiers 272ms
  fetch_shielded_by_tags       191ms
```

A third of the sync phase re-read rows it had already seen and dropped, growing with history — which is why sync went from 714ms to 6244ms across one day on unchanged wallets.

| phase | before | after |
|---|---|---|
| `fetch_proofless_deposits` | 909ms | 275ms |
| `fetch_shielded_by_tags` | 191ms | 113ms |
| `fetch_shielded_by_nullifiers` | 272ms | 180ms |
| `sync_wallet_total` | 2650ms | 1376ms |

The other two streams improved because they no longer queue behind the proofless scan.

`wallet.proofless_cursors` mirrors `sync_cursors` and is kept separate on purpose: they are positions in different streams, and sharing one cursor would skip rows in whichever is behind. Grouping by resume point and advancing every tag in a chunk to the furthest position is the rule the tag stream already uses, sound for the same reason — the ordering is global.

A `proofless_only` flag on `get_encrypted_utxos_by_tags` was rejected: server-side filtering cuts the rows returned but not the pages read, so the client would still walk the whole stream every sync. The cursor removes the reads themselves.

Not yet covered by a test: `MockIndexer` always returns `next_cursor: None`, so pinning this needs a mock that advances, and its 18 construction sites take the field list positionally. Worth doing before merge.

## 2. Make indexer failure a type, not a string to search

`IndexerUnavailable` carried a `String`, so anything needing to know why searched it. The load generator separated throttling from real failure with:

```rust
if lowered.contains("403") || lowered.contains("429") || lowered.contains("rate")
```

That counts a signature containing `429` as a rate limit, counts the word "generate" as one, misses a 503 entirely, and changes meaning whenever an upstream message is reworded.

It now carries an `Unavailable` reason — `Timeout`, `Connect`, `RateLimited`, `ServerError { status }`, `JsonRpcInternal` — plus the transport's text as `detail`, for humans only. `classify()` and `should_retry` match on the reason. Every reason is retryable today, which the enum states rather than assumes: adding a non-retryable one forces `should_retry` to be revisited.

`Timeout` is its own variant because of a specific confusion. reqwest renders a timeout that fires while reading the response body as "error decoding response body", which reads like a deserialization bug. Hundreds of these across a day of load tests were written off as indexer flakiness before anyone noticed `is_timeout()` was true and the client's own read deadline was the cause.

## 3. Group load-test failures by error variant, not by message prefix

Grouping keyed on the first 160 characters of the rendered error, which is not an identity. It failed in both directions in one run: an on-chain rejection arrives with two different wrappers depending on whether it failed in simulation or on send, so 10748 occurrences grouped under a key naming none of them, while 318 occurrences of the same error grouped separately as `0x1b60` because their wrapper was shorter. Widening the cut moves the boundary; it does not make a prefix into an identity.

The key is now the `ClientError` variant's discriminant plus, for a chain rejection, the code from `program_error_code()`. Two failures share a key exactly when they are the same failure. The stored message is one example per group, shown to a reader and used for nothing else.

`ShieldedPoolError` gained a `from_code` reverse lookup, generated with the enum from one list by `macro_rules!` rather than kept as a second hand-maintained table. That matters here: `0x1b60` was read off a `TreeError` conversion and called `StaleNullifierRoot` in three commit messages. It is 7008, `TransactProofVerificationFailed`; `StaleNullifierRoot` is 7015 and never fired. strum's `EnumIter` would generate the same thing, but this crate compiles into the on-chain program and keeps its dependencies to what the program needs.

```
3145x  program error 0x1b5a (nullifier tree maintenance failed)
 177x  rpc error: get_account ... AccountNotFound
  66x  prover server error: async proof failed ... Job timed out in processing
  49x  program error 0x1b60 (transact proof verification failed)
```

Five tests, including the two-wrapper case and the converse: per-occurrence detail such as slots and attempt counts cannot split one cause across many lines.

## 4. Bound the async proof poll by wall clock, and cap the status request

A 220-worker run with a 420s duration sat wedged for 35 minutes. Sampling put all 197 live workers in one frame — `ProverClient::poll_async`, inside a status GET — while the prover's transfer queue had been empty for 20 minutes and photon was answering in 150ms. Nothing was pending; the client could not stop waiting.

`max_wait_secs` was described as a ceiling but compared against a running total of time slept between polls. Time inside a request added nothing to it, so a poll that took a minute counted as zero and the loop went round again with its budget intact. It is now measured with `Instant` from the first poll, so a configured 1200s means 1200s.

The status poll also inherited the 600s request timeout the client sets for `POST /prove`, where a cold prove can take minutes. `GET /prove/status` reads one Redis key. Sharing the prove timeout let a single hung poll hold a worker for ten minutes, and with the accounting above the delays stacked without limit. Status polls now carry their own 30s timeout.

The regression test uses a mock that answers slowly rather than not at all — the case a fast mock cannot reach — and holds the port open afterwards, so polls the client should not have made are recorded instead of hitting a closed socket. Under the old accounting it does not terminate.

This bounds the client. The server-side cause is #210.

`sdk-libs/client`'s `merge_proving` suite fails locally: 2 tests, ~10 minutes. Whether that predates this branch is unestablished — a clean-tree comparison exceeded the time available, and the suite spawns a prover from `target/debug/zolana`. Nothing here touches proving; CI settles it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9843,6 +9843,7 @@ dependencies = [
  "solana-loader-v3-interface 8.0.1",
  "solana-message",
  "solana-pubkey 4.2.0",
+ "solana-rpc-client-api",
  "solana-signer",
  "solana-transaction",
  "swap-program",
--- a/program-libs/interface/src/error.rs
+++ b/program-libs/interface/src/error.rs
@@ -20,19 +20,6 @@ pub enum InterfaceError {
 }
 
 /// Declares the error enum and its reverse lookup from one list.
-///
-/// A client sees only a number -- `custom program error: 0x1b60` -- so naming
-/// it needs a code-to-variant map. Maintaining that map by hand is how 0x1b60
-/// got called `StaleNullifierRoot` for a day: it is 7008,
-/// `TransactProofVerificationFailed`, while `StaleNullifierRoot` is 7015. The
-/// wrong name came from reading a `TreeError` conversion instead of this table
-/// and survived into three commit messages because nothing checked it.
-///
-/// Generating both directions from one list means a new variant cannot be
-/// added without `from_code` learning it. A derive macro (strum's `EnumIter`)
-/// would do the same, but this crate compiles into the on-chain program and
-/// keeps its dependency list to what the program needs; a `macro_rules!` costs
-/// nothing at all.
 macro_rules! shielded_pool_errors {
     ($(
         $(#[$attr:meta])*
@@ -202,11 +189,6 @@ mod tests {
     use super::{ShieldedPoolError, ShieldedPoolError::*};
 
     /// Every pinned code must map back to the variant it came from.
-    ///
-    /// `expected_code` below is exhaustive with no wildcard, so a new variant
-    /// stops this file compiling until it is listed there -- and because the
-    /// enum and `from_code` are generated from one list, the lookup learns it
-    /// at the same moment. This asserts the two really do agree.
     #[test]
     fn every_code_maps_back_to_its_variant() {
         for error in [
--- a/program-libs/interface/src/error.rs
+++ b/program-libs/interface/src/error.rs
@@ -19,14 +19,52 @@ pub enum InterfaceError {
     AlreadyInitialized,
 }
 
-/// Program errors surfaced on-chain as `ProgramError::Custom(code)`.
+/// Declares the error enum and its reverse lookup from one list.
 ///
-/// The discriminants below are the on-chain error codes for this program
-/// version. `error_codes_are_stable` pins the mapping so intentional ABI
-/// changes are explicit.
-#[derive(Clone, Copy, Debug, Error, PartialEq, Eq)]
-#[repr(u32)]
-pub enum ShieldedPoolError {
+/// A client sees only a number -- `custom program error: 0x1b60` -- so naming
+/// it needs a code-to-variant map. Maintaining that map by hand is how 0x1b60
+/// got called `StaleNullifierRoot` for a day: it is 7008,
+/// `TransactProofVerificationFailed`, while `StaleNullifierRoot` is 7015. The
+/// wrong name came from reading a `TreeError` conversion instead of this table
+/// and survived into three commit messages because nothing checked it.
+///
+/// Generating both directions from one list means a new variant cannot be
+/// added without `from_code` learning it. A derive macro (strum's `EnumIter`)
+/// would do the same, but this crate compiles into the on-chain program and
+/// keeps its dependency list to what the program needs; a `macro_rules!` costs
+/// nothing at all.
+macro_rules! shielded_pool_errors {
+    ($(
+        $(#[$attr:meta])*
+        $variant:ident = $code:literal,
+    )*) => {
+        /// Program errors surfaced on-chain as `ProgramError::Custom(code)`.
+        ///
+        /// The discriminants below are the on-chain error codes for this program
+        /// version. `error_codes_are_stable` pins the mapping so intentional ABI
+        /// changes are explicit.
+        #[derive(Clone, Copy, Debug, Error, PartialEq, Eq)]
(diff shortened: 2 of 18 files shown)


sergeytimoshin:
**This can no longer be rebased and merged as it stands.** Against current `main` it is a net −2533 lines, and much of that is reverting work that landed after it:

- deletes `sdk-libs/ts/bench/loadtest.ts` (439 lines), added by #224
- deletes five prover test files (`dedup_cache_test.go`, `processing_queue_test.go`, `dispatch_stage_test.go`, `proof_requests_counted_once_test.go`, `status_lookup_test.go`)
- removes 103 lines from `services/photon/src/api/method/rings/common.rs`
- removes TS wallet-state and wallet-sync test code

Three of its commits are also already on `main` in some form (#199, #200), and `sdk: remember where the proofless scan stopped` (`2a04cc43`) is now in #220, cherry-picked with the resume test its own message said it could not write.

Splitting the parts that are genuinely unlanded instead. First one is #226 — typed indexer failures, error-code grouping in the load-test report, and `xtask fund`. Five commits, cherry-picked onto `main` cleanly, no sync changes, 52 tests green.

Still to split out:
- `client: bound the async proof poll by wall clock`
- `client: fetch spend and dummy nullifier proofs concurrently`
- `client: fetch the blockhash after proving, not before`

One commit needs a decision rather than a move. `2d9198d8 sdk: query chunks concurrently and stop on a short page` does what #220 tried and reverted:

```rust
if last_page || cursor.is_none() {
```

Ending a read on a short page makes the non-advancing-cursor guard unreachable, and stakes correctness on every endpoint returning a short page only when genuinely out of rows — three tests caught it. The concurrency half of that commit is still worth having; the short-page half should not come back.

Suggest closing this once the remaining pieces are split out, rather than rebasing.

sergeytimoshin:
Closing — fully mined into three PRs rather than rebased, since against current `main` this branch is a net −2533 lines and much of that reverts work that landed after it (#224's TypeScript loadtest, five prover test files, photon's `rings/common.rs`).

Where each part went:

| commits | outcome |
| --- | --- |
| `sdk: run a sync round's three queries concurrently` | already on `main` (#200) |
| `client: collect queued proofs on a sub-second backoff` + poll-ceiling follow-ups | already on `main` (#199) |
| `sdk: remember where the proofless scan stopped` | #220, with the resume test this branch said it could not write |
| typed indexer failures, error-code grouping, `fund` | #226 — `fund` folded into `loadtest --funder` rather than kept as its own command |
| bounded proof poll, concurrent proof fetch, blockhash after proving | #227 |

Not carried over: `2d9198d8 sdk: query chunks concurrently and stop on a short page`. Its concurrency half is worth having, but the short-page half is the rule #220 tried and reverted — ending a read on a page short of the limit makes the non-advancing-cursor guard unreachable, and stakes correctness on every endpoint returning a short page only when genuinely out of rows. Three tests caught it. Worth reopening as its own PR with that hunk dropped.

PR #208, opened 2026-08-09
indexer: sort the tag query by the table it filters (NOT VALIDATED — do not merge)

sergeytimoshin:
Not ready to merge. The migration is applied on devnet but the image was reverted, because the evidence supporting this change did not match the query the client actually sends. Opened so the work and the reasoning are not lost.

`get_encrypted_utxos_by_tags` filters on `rings_outputs.view_tag` but orders by `rings_transactions.slot, signature, event_index`. No index spans a join, so Postgres walks `rings_transactions` in slot order and probes outputs per row, discarding nearly everything. At the time, Performance Insights attributed 17.16 of ~17.6 average active sessions to this one statement, with RDS at 98.6% CPU while photon itself used 35–60%.

The branch does three things:

1. Denormalizes `signature` and `event_index` onto `rings_outputs` — the trade `rings_outputs.slot` already makes — with a backfill and `idx_rings_outputs_view_tag_order` on `(view_tag, slot, signature, event_index, output_index)`.
2. Points the ORDER BY and the cursor at `po.*`, so filter and sort come from one index.
3. Rewrites the cursor predicate as a row comparison, `(a, b, c, d) > (w, x, y, z)`, which Postgres can start an index scan at. The equivalent OR chain cannot: the plan read from the tag's first row and discarded everything before the cursor (`Rows Removed by Filter: 1201`), so each page cost more than the last.

Single-tag measurements, with the real cursor predicate:

| | execution | buffers |
|---|---|---|
| original | 1.854ms | 3440 |
| + denormalize & index | 0.891ms | 1234 |
| + row comparison | 0.434ms | 662 |

And flat with depth: a page 2200 rows in cost 0.487ms against 0.434ms at 1200 in.

Every one of those EXPLAINs used one `view_tag`. The client sends 64 (`DEFAULT_TAG_QUERY_CHUNK`). Those are different queries with different plans, and the wrong one was validated — three times. With 64 literal tags the gap is much smaller:

| 64 tags, with cursor | execution | buffers |
|---|---|---|
| current ordering | 0.728ms | 1191 |
| this branch | 0.655ms | 431 |

Better on buffers, roughly a wash on time.

The devnet run afterwards gave 6.34 tps against 7.66 before, but that comparison is confounded: `first sync mean` went from 3396ms to 6877ms between the two runs, so wallet depth roughly doubled, which is what the bench warns invalidates a cross-run comparison. Bloat was ruled out separately — `n_dead_tup = 0`, and an explicit `VACUUM (ANALYZE, VERBOSE)` removed nothing.

No demonstrated benefit, no ruled-out harm. It went back rather than shipping on evidence that did not match the workload.

The database was subsequently moved from 2 to 8 vCPU, which dropped RDS CPU from ~99% to ~8% and removed the pressure this was meant to relieve. Before reviving this branch:

1. EXPLAIN both orderings with 64 literal tags — dump them and inline them, since `IN (SELECT ...)` becomes a semi-join and changes the plan.
2. Compare runs at similar wallet depth, or the throughput numbers mean nothing.
3. Decide whether denormalized columns are worth carrying at all now that the instance is not saturated.

State on devnet: the migration is applied. `signature` and `event_index` exist on `rings_outputs`, backfilled, with the index built and verified — 0 rows missing the ordering key, 0 mismatched against the source columns. The running image does not reference them, so they are inert; the only cost is the extra write work maintaining the index.


sergeytimoshin:
Back to draft. Marking this ready was a mistake on my part — the title and body both say the evidence was invalidated, and the full suite then found it does not compile: the denormalized `signature` and `event_index` columns were added to the model but not to the rings fixtures, which fails `cargo check`, clippy, coverage and the photon tests.

Not worth fixing the fixtures until the questions in the body are settled: re-EXPLAIN with 64 literal tags, compare runs at similar wallet depth, and decide whether the columns are worth carrying now that RDS runs at ~8% CPU rather than ~99%.

sergeytimoshin:
Superseded by #223: same commits rebased onto `main`, with the rings migration-list conflict resolved. Closed because GitHub's stack feature would not let this PR's base move off `indexer/cursor-watermark`, which is already merged.

PR #210, opened 2026-08-09
prover: answer a status poll with two key reads instead of four queue scans

sergeytimoshin:
Stacked on #202. Review that first.

`checkJobExistsDetailed` runs on every `/prove/status` request — the hottest path in the server, since a client polls it from submission until its proof lands — and answered by walking whole Redis lists with `LRANGE key 0 -1`: the job's own queue, then the processing queue, then `zk_failed_queue`, then `zk_results_queue`, unmarshalling every element of each.

A 220-worker load test made it visible. In one ten-minute window:

| | |
|---|---|
| status polls | 34,154 (~57/s) |
| `zk_results_queue` length | 8,237 |
| implied element reads | ~470k/s from a single-threaded Redis |
| Redis `i/o timeout` | 221 |
| `context deadline exceeded` | 81 |

Dispatch stalls behind the same Redis, health checks fail, ECS replaces the tasks, and the jobs in flight on them are orphaned. Throughput held ~24.5 tps for four consecutive 30s windows, then went to 0.00 and stayed there.

This is the fourth instance of the same shape — a linear Redis list scan on a per-job path — after #202 and its two parents. It mattered most, because it is per poll rather than per job.

The scans were there because job metadata was written once at submission with `status: "queued"` and never updated, so `queued` and `processing` could only be told apart by looking in both queues. Failures were worse: the metadata was deleted on failure, leaving `zk_failed_queue` as the only record, so every poll for an unfinished job scanned that too.

The workers now keep the metadata current:

- `MarkJobProcessing` when a job enters the processing queue
- `MarkJobFailed` on every terminal path — proof error, panic, expiry, replayed cached failure, and the stuck-job reaper — carrying the failure details that used to be read back out of the queue entry

That is one `SET` per job against one scan per poll, and polls outnumber jobs by orders of magnitude. The status path is now `GetResult`, already O(1), plus one metadata `GET`.

The reaper case bites if missed: it fails jobs whose metadata may already have expired, so `MarkJobFailed` recreates the record rather than skipping it. Without that, a reaped job is indistinguishable from an unknown one and the client waits out its entire deadline — which is how 197 workers ended up parked for 35 minutes against an idle prover.

`findJobInQueue` and `extractResultFromPayload` are removed along with their last callers.

`TestFailedJobStatusHTTPEndpoint` built its fixture by writing straight into `zk_failed_queue`, which only worked because of the scan; it now fails the job the way the workers do. Five new tests cover the transitions, that picking a job up preserves its submission expiry rather than renewing it, and that a reaped job still leaves a terminal record.

`go build`, `go vet`, `gofmt` and the full `go test ./...` clean.

This removes the collapse. It does not raise the ceiling: `NullifierTreeUpdateFailed` (7002) was 3,145 of 3,474 failures in the same run and is what caps throughput at ~24 tps. That is separate.


sergeytimoshin:
Superseded by #222: same commits rebased onto `main`. Closed because GitHub's stack feature would not let this PR's base move off `prover/queue-wait-metric`, which is already merged.

PR #211, opened 2026-08-09
forester: overlap proving with witness building, and fix the quadratic low-element lookup

sergeytimoshin:
The forester proved one zkp-batch, submitted it, then built the next witness. Each proof takes about a minute and clears one zkp-batch of nullifiers, so the pool was capped at `zkp_batch_size / proof_seconds` — 4.2 spends/second on devnet, against a client sustaining 24. When the nullifier queue filled, the program rejected spends outright: 3145 of 3474 transfer failures in one 220-worker run were `NullifierTreeUpdateFailed`.

| forester | nullifiers/s |
| --- | --- |
| serial, 6 vCPU | 4.2 |
| serial, 44 vCPU | 14.1 |
| fixed groups of 8, 44 vCPU | 13.5 |
| this branch | 30.3 |

### DIFF
--- a/forester/src/cli.rs
+++ b/forester/src/cli.rs
@@ -1,6 +1,8 @@
 use clap::{Parser, Subcommand};
 use solana_pubkey::Pubkey;
 
+use crate::run::DEFAULT_PROOF_CONCURRENCY;
+
 /// Default SPP pool tree the forester maintains (devnet/localnet fixture).
 pub const DEFAULT_TREE: Pubkey = Pubkey::from_str_const(zolana_interface::DEFAULT_TREE_ADDRESS);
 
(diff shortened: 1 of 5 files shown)


sergeytimoshin:
**The do-not-merge note at the top is now stale — the regression is fixed.**

This PR was held because out-of-order submission wedged photon: the program applies several zkp batches under one cascade event, and the indexer recorded the instruction's claimed root instead of the event's. That is fixed in #212, along with two further defects it exposed (both trees derived their client-facing root index from a locally counted sequence that had drifted from the chain's ring pointer).

Verified together on devnet with `--proof-concurrency 8`: **16.35 tps sustained, 6,950 transfers, zero failures** over 425s at 80 workers, with zero indexer errors.

The forester side of that run: ~250 nullifiers cleared every 7s (~35/s) against 16.4 tps demand, never more than 2 batches queued. So the parallel proving in this PR is doing its job and the forester is no longer the bottleneck — proving latency in the client loop is (p50 2,524ms of a 4,655ms total).

Depends on #212. Once that lands this is mergeable; I have not edited the PR body, so the warning there should be removed before merge.

ananas-block (approved):

PR #212, opened 2026-08-09
photon: index the batch append event, not the instruction that requested it

sergeytimoshin:
Unblocks #211, which wedges the indexer without it.

`BatchUpdateNullifierTree` is a request, not a record. The program verifies the proof, caches the update at its zkp batch index, then applies whatever has become contiguous. One instruction can therefore apply nothing — its proof arrived before an earlier one and waits — or apply several zkp batches at once, when it unblocks the ones cached behind it. Only the emitted `BatchAddressAppendEvent` says which batches landed and what root the tree took.

Photon read the instruction, took the forester's claimed `new_root`, appended one zkp batch of queued nullifiers, and reconstructed something else:

```
Failed to index block batch 482473523-482473523.
Parser error: Reconstructed nullifier root mismatch for tree trEE...
```

It retried that block indefinitely. devnet sat at 10% CPU with lag growing from 4697 to 6109 slots, and a 220-worker loadtest completed 0 transfers.

Nothing was corrupted. `verify_reconstructed_root` runs inside the ingest transaction and returns `Err`, so the bad root was never committed. Recovery is a restart with `--start-slot <last indexed>`.

The program's own comment states the contract the parser did not honour: *"Photon's parser records batch updates from the emitted event in successful transactions only."*

The parser now reads the event. `num_update` × `zkp_batch_size` is how many queued nullifiers the cascade consumed, and the root is verified once, against the root after all of them. The guard stays: it is what made this a loud stall rather than silent corruption. An instruction that applied nothing emits nothing, and is now ignored rather than recorded.

`BatchAddressAppendEvent` moves to `zolana-event`, beside the `EventKind` variant that names it and the encoder that emits it. photon already depends on that crate; the alternative was linking the whole batched tree library — bloom filters and groth16 verifier included — for one struct.

Event discovery moves to `parser/event_site.rs`, shared with the rings parser and parameterized by which parent tags count as a source. An `EMIT_EVENT` is an inner instruction whose payload any program can forge by CPI; its parent cannot be forged. That check now has one implementation rather than two.

Any cascade triggers this, not only our forester: a retried transaction, two batches landing in one block, or a second forester implementation. It was reachable before #211.

66 tests pass. The load-bearing one is `cascade_lands_the_same_tree_as_the_batches_applied_one_at_a_time` — three batches under one event must leave exactly the tree that three separate events leave, which is the property the old code lacked. Also covered: a cascade missing part of its queued nullifiers fails rather than applying what it has; an instruction that emitted no event contributes nothing; and the forged-parent cases move with the extracted module, gaining one for a rejected source tag.

`cargo clippy --workspace --all-targets` and `cargo fmt --all` clean.

Deploy: rebuild photon, restart with `--start-slot <last indexed slot>`, then raise `forester_proof_concurrency`.


sergeytimoshin:
Second commit, from deploying this to devnet: the cascade fix was incomplete and the gap was invisible in tests.

The API serves `root_seq % root_history_capacity` as the root index a client quotes. On chain that index is a ring pointer advanced **once per applied zkp batch**. Photon counted its own sequence, once per batch update ingested — which matched exactly as long as one instruction applied one batch, and stops matching the moment a cascade applies several under one event.

I reasoned about this in the first commit and got it wrong, concluding photon's `seq` was "only used for row ordering and upsert conflict resolution, never exposed as a chain sequence number." It is exposed, as the root index.

Measured on devnet after deploying the first commit:

| | chain | photon |
|---|---|---|
| sequence_number | 721 | 718 |
| root index for the served root | 1 | 118 |

The root *value* photon served was correct and present on chain — the tree reconstructs fine. Only the index was wrong, so clients quoted history slot 118 for a root living at slot 1, the program verified against an unrelated root, and every transfer failed `TransactProofVerificationFailed` (7008). 112 of 112 in an 8-worker run, zero successes.

The event already carries the chain's number, so photon should not be counting at all: `start_sequence_number + num_update - 1`. That also repairs an already-drifted index on the next update instead of needing a reindex. `next_root_sequence` is deleted.

Both new tests fail against the previous commit — verified by mutation, not assumed.

sergeytimoshin:
Verified end-to-end on devnet. **16.35 tps sustained, 6,950 transfers, zero failures, zero throttling** over 425s at 80 workers — the first clean run of the full pipeline.

Root index now matches the chain exactly, checked by decoding the account rather than trusting the API:

| | value |
|---|---|
| chain `root_history_cursor` | 155 |
| `root_history[155]` | equals the `root` field |
| photon serves | **155** (was 175) |

photon logged zero mismatches, stale roots or internal errors across the run. The forester cleared ~250 nullifiers every 7s (~35/s) against 16.4 tps demand, never queueing more than 2 batches, so it is no longer the constraint.

Where the time goes now — throughput is exactly the latency loop (80 workers ÷ 4.66s ≈ 17 tps):

| phase | p50 |
|---|---|
| prove | 2,524ms |
| send | 1,105ms |
| sync | 634ms |
| confirm | 257ms |
| **total** | **4,655ms** |

One measurement caveat worth recording: the earlier 220-worker collapse was **wallet history**, not indexer capacity. First sync was 35,770ms on wallets reused all session versus 2,316ms on fresh ones. Any future comparison needs fresh wallets — the loadtest's own warning about this is accurate.

ananas-block (approved):

PR #213, opened 2026-08-10
sdk: require https for indexer and prover, and add a TypeScript loadtest

sergeytimoshin:
Two commits, found by building the TypeScript loadtest and discovering it could not talk to devnet at all.

Based on `sdk/proofless-cursor` (#207), where `xtask loadtest` lives.

The SDKs disagreed on a security property. The TypeScript SDK refuses plain `http://` to non-loopback hosts; the Rust client accepted anything. So the Rust loadtest has been talking to devnet in the clear, and no TypeScript integrator can reach devnet at all.

What these endpoints carry is the reason it matters:

- the indexer response is which UTXOs an identity owns
- the prover request carries the witness

In plaintext both are readable by anyone on the path — the privacy property the protocol exists to provide, not a hardening detail. A tamperer can also serve altered merkle proofs or root indices and steer a client into proving against state it did not choose.

`ZolanaClient::from_urls` now validates and returns a `Result`, with `from_urls_allowing_insecure_http` as the escape hatch for a transport that is already private: indexer inside a VPC, TLS terminated elsewhere. It is a named function rather than a boolean so the insecure choice is greppable. TypeScript gains the matching `allowInsecureHttp`, threaded through `ProverClient`, which re-validates independently.

Both loadtests set the opt-in explicitly, because devnet's ALB has no certificate (`acm_certificate_arn = ""`). That exposure now appears in the code that depends on it rather than being the silent default, and it goes away when the certificate is issued.

Tests cover the rule in both SDKs, including the lookalike hosts a `127.` prefix check must reject — `127.0.0.1.evil.com`, `localhost.evil.com` — and that the opt-in relaxes the scheme only, still rejecting credentials and fragments.

The second commit adds `npm run loadtest`, mirroring `xtask/src/loadtest.rs` closely enough that the numbers compare: same wallet-file format, same ring of transfers, same four phases, same 30s windows, same typed error grouping. The shielded derivation matches `ShieldedKeypair::from_solana_keypair`, so wallet directories funded by `xtask fund` work unchanged. It runs against `dist/`, measuring the package as published, and adds no dependency since Node strips the types natively.

First comparison, 8 workers, same fresh wallets, same devnet:

| phase p50 | Rust | TypeScript |
|---|---|---|
| sync | 517ms | 2,683ms |
| prove | 1,349ms | 4,811ms |
| send | 1,079ms | 566ms |
| confirm | 226ms | 1,090ms |
| total | 3,126ms | 9,258ms |
| tps | 2.56 | 0.82 |

Proving is the gap worth chasing: the same prover service serves both, so 3.5x is client-side assembly rather than the backend. Sync is inflated partly by wallet history, which grows every run, so the harness reports first-sync cost — runs are only comparable at similar depth. The prove gap is not investigated here; this PR is the instrument, not the fix.

`cargo test -p zolana-client --features client` 53 pass; TypeScript 230 pass across 15 files; typecheck, oxlint and oxfmt clean. The client crate's tests need `--features client` — without it the module is cfg'd out and `cargo test -p zolana-client` silently runs a different 25.

### DIFF
--- a/cli/src/wallet_cli/transaction.rs
+++ b/cli/src/wallet_cli/transaction.rs
@@ -28,7 +28,7 @@ pub(crate) fn run_transfer(opts: TransferOptions) -> Result<()> {
         network.sync.indexer_url.clone(),
         network.prover_url.clone(),
         Address::new_from_array(network.tree.to_bytes()),
-    );
+    )?;
     let recipient = parse_pubkey(&opts.to)?;
 
     let transfer = create_transfer_sync(TransferParams {
@@ -111,7 +111,7 @@ pub(crate) fn run_split(opts: SplitOptions) -> Result<()> {
         network.sync.indexer_url.clone(),
         network.prover_url.clone(),
         Address::new_from_array(network.tree.to_bytes()),
-    );
+    )?;
     let input = opts
         .input
         .as_deref()
@@ -183,7 +183,7 @@ pub(crate) fn run_merge(opts: MergeOptions) -> Result<()> {
         network.sync.indexer_url.clone(),
         network.prover_url.clone(),
         Address::new_from_array(tree.to_bytes()),
-    );
+    )?;
     // Submit only needs the owner's public identity plus the nullifier secret;
     // no signing/viewing/funding secret crosses the submit boundary.
     let material = MergeMaterial::from_keypair(&ctx.material.keypair);
--- a/cli/src/wallet_cli/withdraw.rs
+++ b/cli/src/wallet_cli/withdraw.rs
@@ -29,7 +29,7 @@ pub(crate) fn run_withdraw(opts: WithdrawOptions) -> Result<()> {
         network.sync.indexer_url.clone(),
         network.prover_url.clone(),
         Address::new_from_array(network.tree.to_bytes()),
-    );
+    )?;
     let recipient = parse_pubkey(&opts.to)?;
 
     // An SPL withdrawal settles into the recipient's associated token account,
--- a/sdk-libs/client/src/client.rs
+++ b/sdk-libs/client/src/client.rs
@@ -7,6 +7,48 @@
 
 use std::{sync::OnceLock, thread::sleep, time::Duration};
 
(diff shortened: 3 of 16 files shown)


sergeytimoshin:
Instrumented the prove phase and profiled it. The 3.5x gap is real and it is not the hasher.

**Phases are comparable** — I checked before drawing conclusions. Rust brackets `create_transfer_sync` + `sign_private_transaction_sync`, which includes the same two indexer round-trips and the same prover call the TS `prove` covers.

**Prove splits as:**

| | p50 |
|---|---|
| prove total | 4,818ms |
| network (indexer + prover) | 1,573ms |
| **local compute** | **3,080ms** |

Local alone is more than twice the entire Rust prove phase (1,349ms, network included).

**Not Node's single thread**: at 2 workers local is 3,002ms versus 3,080ms at 8 — unchanged, so it is per-transfer cost, not queueing.

**CPU profile**, of active (non-idle) time:

| | share |
|---|---|
| `@noble/curves` modular + weierstrass arithmetic (pure JS) | ~68% |
| Poseidon WASM (`light_wasm_hasher`) | ~8% |

So it is **P-256 scalar multiplication in JavaScript**. `ecdhX` calls `p256.getSharedSecret` once per output when building, and wallet sync calls it once per candidate output it tests for ownership — which is why prove *and* sync are both slow, and why sync grows with wallet history.

Options, roughly in order of effort:

1. **WASM P-256.** The package already ships a Rust-compiled WASM hasher; extending that to curve ops would close most of the gap. Biggest win, biggest change.
2. **Cache the shared secret per counterparty.** Repeated sends to the same recipient recompute the identical ECDH today. Legitimate for real wallets, not just the benchmark.
3. **noble precomputation** (`precompute` / window size) for repeated points.

I have not implemented any of these — this PR is the instrument and the diagnosis.

sergeytimoshin:
Superseded by #224: same commits rebased onto `main`, no longer stacked on #207. Closed because GitHub's stack feature would not let this PR's base move off `sdk/proofless-cursor`.

PR #214, opened 2026-08-11
ci: publish photon, the prover and the forester from one workflow

sergeytimoshin:
Publishes photon, the prover, and the forester from one workflow. `photon-image.yml` becomes `publish-image.yml`.

Images are built in the repository that owns the code. The prover and forester were previously built in the infra repository, which checked out this source: a change here could break a build there, and publishing required driving the other repository.

Two channels:

- `release` — the tag must be `<service>-zolana-<sha12>`, and the commit must be an ancestor of `main`.
- `preview` — any branch, any tag.

Both are immutable and attested. `release` is the rule `photon-image.yml` already applied to photon. `preview` covers branch builds, which is how every image currently deployed was produced.

Registry is `558215002830.dkr.ecr.eu-north-1.amazonaws.com`. The previous target, `public.ecr.aws/f7o9l7p1/photon`, holds upstream Helius photon — different code. Nothing was ever published there: the job was gated on a tag never cut, and `vars.ECR_HELIUS_PROD_AWS_ROLE_ARN` was unset.

The OIDC subject is `repo:helius-labs/zolana:environment:image-publish`. This repository predates GitHub's immutable-subject default, so it uses plain `owner/repo`; declaring `environment:` replaces the `:ref:` segment.

Each published digest carries a provenance attestation:

```
gh attestation verify oci://<image> --owner helius-labs
```

The prover builds from `prover/server/Dockerfile.light`, which ships no proving keys; they are fetched at runtime and verified against the lockfile embedded in the binary. `prover/server/Dockerfile` bakes them, adding 1.4–3.9 GB.

Verified: prover and forester published through `preview`; attestation recorded and subject confirmed; photon's validate job skipped for non-photon dispatch; the photon tag-push path unchanged.

The infra repository's build path is removed in parallel (helius-labs/zolana-infra), leaving only the dashboard aggregator, which is built from that repository.

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -1,4 +1,31 @@
-name: photon-image
+# Publishes the service images to our own ECR: photon, the prover, the forester.
+#
+# These live here, not in the infra repository, because this is the repo that
+# owns the code. Building them from infra meant infra CI checked out this
+# source, so a change here could break a build there, and a protocol change
+# could not publish an image without driving the other repository.
+#
+# Two channels, because one rule could not fit both jobs:
+#
+#   release  the tag must name the commit (<service>-zolana-<sha12>) and the
+#            commit must be on main. This is photon's original rule, now
+#            available to every service.
+#   preview  any branch, any tag. Still immutable, still attested; it just does
+#            not claim to be a release.
+#
+# Preview exists because the release rule would ban what we actually do -- every
+# image running on devnet today was built from a feature branch -- and a rule
+# that bans normal work gets bypassed rather than followed. That is what
+# happened before: photon had this gate, and every photon image ever deployed
+# was built by a different pipeline that had none.
+#
+# The aggregator is NOT here. It is built from the infra repository's dashboard/
+# directory and is not a protocol service.
+#
+# Requires:
+#   vars.AWS_IMAGE_PUBLISHER_ROLE_ARN  image_publisher_role_arn from infra/bootstrap
+
+name: publish-image
 
 on:
   push:
@@ -14,7 +41,7 @@ on:
       - "rust-toolchain.toml"
       - "LICENSE"
       - "THIRD_PARTY_NOTICES"
-      - ".github/workflows/photon-image.yml"
+      - ".github/workflows/publish-image.yml"
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     paths:
@@ -27,11 +54,25 @@ on:
       - "rust-toolchain.toml"
       - "LICENSE"
       - "THIRD_PARTY_NOTICES"
-      - ".github/workflows/photon-image.yml"
+      - ".github/workflows/publish-image.yml"
   workflow_dispatch:
     inputs:
+      service:
+        description: Which image to publish
+        required: true
+        default: photon
+        type: choice
+        options: [photon, prover, forester]
+      channel:
+        description: >-
+          release enforces the commit-naming tag and requires the commit to be
+          on main; preview allows a branch build under any tag
+        required: true
+        default: preview
+        type: choice
+        options: [preview, release]
       image_tag:
-        description: Immutable fork tag, photon-zolana-<12-character Zolana commit>
+        description: Tag to publish. For release, <service>-zolana-<12-char commit>
         required: true
         type: string
 
@@ -47,21 +88,24 @@ env:
   # different code, binaries in /app, running as root -- so publishing the fork's
   # image there would have put two unrelated things behind one name. This
   # workflow never actually published, so nothing was ever pushed there.
-  IMAGE_NAME: 558215002830.dkr.ecr.eu-north-1.amazonaws.com/zolana-photon
+  REGISTRY: 558215002830.dkr.ecr.eu-north-1.amazonaws.com
   AWS_REGION: eu-north-1
-  ECR_REPOSITORY: zolana-photon
 
 jobs:
   validate:
     name: container / photon
     runs-on: ubuntu-24.04
-    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    # Photon's own image check. The prover and forester are built and tested by
+    # prover-server.yml and forester.yml, so this does not gate them.
+    if: >-
+      (github.event_name != 'pull_request' || github.event.pull_request.draft == false) &&
+      (github.event_name != 'workflow_dispatch' || inputs.service == 'photon')
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           fetch-depth: 0
       - name: Validate release source
-        if: startsWith(github.ref, 'refs/tags/photon-zolana-') || github.event_name == 'workflow_dispatch'
+        if: startsWith(github.ref, 'refs/tags/photon-zolana-')
         run: |
           git fetch --no-tags origin +refs/heads/main:refs/remotes/origin/main
           if ! git merge-base --is-ancestor "$GITHUB_SHA" origin/main; then
@@ -95,13 +139,15 @@ jobs:
           test "$(docker run --rm --entrypoint id zolana-photon:ci -u)" = 10001
 
   publish:
-    name: publish / photon
+    name: publish / ${{ inputs.service || 'photon' }}
     runs-on: ubuntu-24.04
     needs: validate
     if: >-
-      startsWith(github.ref, 'refs/tags/photon-zolana-') ||
-      github.event_name == 'workflow_dispatch'
-    environment: photon-production
+      always() && needs.validate.result != 'failure' && (
+        startsWith(github.ref, 'refs/tags/photon-zolana-') ||
+        github.event_name == 'workflow_dispatch'
+      )
+    environment: image-publish
     permissions:
       contents: read
       id-token: write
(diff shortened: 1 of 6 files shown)


ananas-block:
is this an issue? 
image

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+          echo "channel=$CHANNEL" >> "$GITHUB_OUTPUT"
+
+      # Release is the strict channel: the tag must name the commit, and the
+      # commit must be on main. Preview exists because that rule would ban what
+      # we actually do -- every image running on devnet today was built from a
+      # feature branch -- and an unenforceable rule gets bypassed rather than
+      # followed. A preview build is still attested and still immutable; it just
+      # does not claim to be a release.
+      - name: Validate the release
+        if: steps.plan.outputs.channel == 'release' && github.event_name == 'workflow_dispatch'
```
ananas-block: A preview dispatch with `image_tag: photon-zolana-<12 hex chars>` skips this step and publishes a tag indistinguishable from a release. Reject release-shaped tags on the preview path:

```bash
if [[ "$IMAGE_TAG" == *-zolana-* ]]; then
  echo "::error::preview tags must not use the release naming scheme" >&2
  exit 1
fi
```

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+          fi
+      - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+      - uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+        with:
+          role-to-assume: ${{ vars.AWS_IMAGE_PUBLISHER_ROLE_ARN }}
+          role-session-name: zolana-photon-publish
+          aws-region: ${{ env.AWS_REGION }}
+      # Private ECR, so no registry-type override. The role is trusted for this
+      # job's OIDC subject specifically -- `repo:helius-labs/zolana:environment:
+      # photon-production`, which is what declaring `environment:` above makes
```
ananas-block: The comment says the role trusts `environment:photon-production`, but the job declares `environment: image-publish`. If the trust policy still says `photon-production`, the first publish fails to assume the role. Fix the comment and confirm the trust-policy subject matches `image-publish`.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ steps.plan.outputs.repository }}
+          flavor: latest=false
+          tags: |
+            type=match,pattern=(photon-zolana-.*),group=1,enable=${{ startsWith(github.ref, 'refs/tags/photon-zolana-') }}
+            type=raw,value=${{ inputs.image_tag }},enable=${{ github.event_name == 'workflow_dispatch' }}
+            type=sha,format=long,prefix=sha-
+          labels: |
+            org.opencontainers.image.title=Zolana Photon
```
ananas-block: These labels apply to all three services, so prover and forester images get `title=Zolana Photon` and the indexer description. Emit `title=`/`description=` from the "Resolve the build" step instead.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+      # photon-production`, which is what declaring `environment:` above makes
+      # GitHub present instead of the ref.
+      - uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64 # v2
+      - id: metadata
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ steps.plan.outputs.repository }}
+          flavor: latest=false
+          tags: |
+            type=match,pattern=(photon-zolana-.*),group=1,enable=${{ startsWith(github.ref, 'refs/tags/photon-zolana-') }}
```
ananas-block: Enabled by ref alone. A dispatch started from a `photon-zolana-X` tag ref with `service: prover` yields two non-`sha-` tags -- `zolana-prover:photon-zolana-X` plus `inputs.image_tag` -- and the publish loop keeps only the last one, silently dropping the other.

```suggestion
            type=match,pattern=(photon-zolana-.*),group=1,enable=${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/photon-zolana-') }}
```

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+        with:
+          context: ${{ steps.plan.outputs.context }}
+          file: ${{ steps.plan.outputs.file }}
+          platforms: linux/amd64
+          load: true
+          push: false
+          tags: ${{ steps.metadata.outputs.tags }}
+          labels: ${{ steps.metadata.outputs.labels }}
+          cache-from: type=gha,scope=photon-image-production
+          cache-to: type=gha,mode=max,scope=photon-image-production
```
ananas-block: One scope now serves three unrelated builds with `mode=max`; they evict each other inside the 10 GB GHA cache budget. Key it per service:

```yaml
cache-from: type=gha,scope=publish-${{ steps.plan.outputs.service }}
cache-to: type=gha,mode=max,scope=publish-${{ steps.plan.outputs.service }}
```

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+          docker run --rm --entrypoint photon-snapshotter zolana-photon:ci --version
+          docker run --rm --entrypoint photon-snapshot-loader zolana-photon:ci --version
+          test "$(docker run --rm --entrypoint id zolana-photon:ci -u)" = 10001
+
+  publish:
+    name: publish / ${{ inputs.service || 'photon' }}
+    runs-on: ubuntu-24.04
+    needs: validate
+    if: >-
+      always() && needs.validate.result != 'failure' && (
```
ananas-block: `needs.validate.result != 'failure'` also passes when validate was cancelled, so a tag-push publish can proceed without the smoke-tested build. Use the positive form: `needs.validate.result == 'success' || needs.validate.result == 'skipped'`.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+            [[ "$tag" != sha-* ]] || continue
+            set +e
+            error="$(aws ecr describe-images \
+              --region "$AWS_REGION" \
+              --repository-name "$ECR_REPOSITORY" \
+              --image-ids "imageTag=$tag" 2>&1)"
+            status=$?
+            set -e
+            if [[ $status -eq 0 ]]; then
+              echo "Refusing to overwrite published Photon tag: $tag" >&2
```
ananas-block: The error messages here and at lines 338, 362, 377 say "Photon" but run for the prover and forester too -- interpolate `$SERVICE`. Same for `role-session-name: zolana-photon-publish` (line 222): CloudTrail would attribute prover/forester publishes to a photon session.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,410 @@
...
+        default: preview
+        type: choice
+        options: [preview, release]
+      image_tag:
+        description: Tag to publish. For release, <service>-zolana-<12-char commit>
+        required: true
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ (startsWith(github.ref, 'refs/tags/photon-zolana-') || github.event_name == 'workflow_dispatch') && 'production' || github.ref }}
```
ananas-block: All dispatches share the `production` group, and GitHub keeps at most one pending run per group -- dispatching prover then forester during a photon publish cancels one of them. ECR enforces immutability now, so keying the group by `inputs.service` is safe.

prover/server/Dockerfile.light
```diff
@@ -1,3 +1,22 @@
...
+# So baking keys is pointless and slightly harmful: the lockfile already makes
+# pk<->vk<->code drift impossible, and a baked key layer is 1.4 GB for client
+# shapes to 3.9 GB for forester batch shapes, rebuilt and re-pulled on every
+# code change. That is what prover/server/Dockerfile does; prefer this one.
+#
+# One image serves both roles. The client prover and the forester prover differ
+# only in flags -- circuit selection and concurrency -- and in which keys they
+# happen to fetch.
+
 FROM golang:1.25-alpine AS builder
```
ananas-block: This PR makes this image published and attested, but the base images (here and line 36, and both in `forester/Dockerfile`) are mutable tags, so the same commit can produce different bytes depending on pull time -- which weakens the provenance attestation. Pin by digest like `services/photon/Dockerfile` does.

CLAUDE.md
```diff
@@ -454,9 +454,10 @@ git tag "$tag"
 git push origin "$tag"
 ```
 
-A manual `photon-image.yml` dispatch must use that same commit-derived value
-for its `image_tag`. The protected `photon-production` environment gates
-publication. The workflow publishes the
+A manual `publish-image.yml` dispatch on the `release` channel must use that
+same commit-derived value for its `image_tag`, and the commit must be on
+`main`; the `preview` channel takes any branch and any tag, for images that are
+not releases. The protected `image-publish` environment gates publication. The workflow publishes the
```
ananas-block: Nit: line overflows the paragraph's wrap width; re-wrap.

ananas-block (commented):
Review of the publish-image workflow. Findings as inline comments, roughly ordered by severity: the preview/release tag namespace guard (1) and the OIDC trust-policy mismatch (2) are the two worth resolving before merge.

sergeytimoshin:
> is this an issue? image

no, it was my test (and its passed)

ananas-block (approved):

PR #217, opened 2026-08-12
Start Photon by default in zolana dev start

tilo-14:
## Goal

`zolana dev start` starts the full localnet (validator, prover, Photon) so installing the CLI is enough to run client examples — no zolana checkout or locally built programs.

## Changes

- Photon starts by default. `--skip-indexer` turns it off, matching `--skip-prover`.
- `cli/README.md` documents the skip flags.

### DIFF
--- a/cli/src/args.rs
+++ b/cli/src/args.rs
@@ -220,7 +220,18 @@ pub(crate) struct TestValidatorOptions {
     #[arg(long, help = "Do not start the prover server")]
     pub(crate) skip_prover: bool,
 
-    #[arg(long, help = "Start a local Photon indexer")]
+    #[arg(
+        long,
+        help = "Do not start the Photon indexer",
+        conflicts_with = "with_photon"
+    )]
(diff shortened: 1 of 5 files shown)


cli/src/args.rs
```diff
@@ -220,7 +220,14 @@ pub(crate) struct TestValidatorOptions {
...
 
-    #[arg(long, help = "Start a local Photon indexer")]
+    #[arg(
+        long,
+        help = "Do not start the Photon indexer",
+        conflicts_with = "with_photon"
+    )]
+    pub(crate) skip_indexer: bool,
+
+    #[arg(long, hide = true, help = "Deprecated alias: Photon starts by default")]
```
ananas-block: please dont deprecate anything as long as we are not live we can make breaking changes in the cli
just remove the legacy option

ananas-block (approved):

sergeytimoshin (approved):

PR #220, opened 2026-08-13
sdk, photon: resume every stream the wallet reads

sergeytimoshin:
Every stream the wallet reads now resumes where it stopped. Tags already did;
nullifiers and proofless deposits did not.

## Changes

- **Nullifiers**: query only unspent notes. A nullifier appears at most once on
  chain, so once its spend is known the answer is final. 293 notes to 56 on the
  measured wallet.
- **Photon**: report `scanned_through`, the position a scan reached. Unspent
  nullifiers match nothing, so `next_cursor` — the last row's position — is
  always absent for them, and no client-side change can supply a resume point.
- **Proofless**: persist per-tag cursors, cherry-picked from #207
  (`2a04cc43`). Its other three concerns are unrelated to sync.
- Both SDKs; `group_by_resume_point` / `groupByResumePoint` shared by all three
  streams.

## Measured

`xtask loadtest`, devnet, two wallets, 90s, unaffordable transfer amount so
wallet state is frozen. Round-0 p50, the three fetches concurrent as of #200:

| phase | main | this PR |
| --- | --- | --- |
| `fetch_shielded_by_nullifiers` | **735ms** | **78ms** |
| `fetch_shielded_by_tags` | 465ms | 455ms |
| `fetch_proofless_deposits` | 803ms | 794ms |
| `sync_wallet_total` | 1051ms | 854ms |

Sync gets 19% faster, not 9x: #200 made a round cost the slowest of its three
queries, and the proofless scan is now that. `sync_wallet_total` max also drops
5594ms → 1958ms.

Devnet runs a photon without `scanned_through`, so that run isolates the
narrowed query set. **Resuming is not yet measured end to end** — it needs a
photon deploy. Unit tests cover it on both sides, including resuming from a
reported position and still finding a spend that landed after it.

## Notes

- **Ship photon and the clients together.** Both schemas are strict, so the
  moment photon sent a field the SDKs did not know, the e2e suite failed every
  nullifier query with `CLIENT_INVALID_RPC_RESPONSE`.
- A read ends when the indexer returns no cursor, never on a short page.
  Stopping early saves a request per chunk but makes the non-advancing-cursor
  guard unreachable.
- Cursors live on the in-memory `Wallet`, so they help a long-lived wallet, not
  successive CLI invocations.
- The tag stream still cannot resume a tag that has never matched anything, for
  the same reason nullifiers could not. Masked because a wallet's own tags
  usually match. Follow-up.

### DIFF
--- a/sdk-libs/transaction/src/wallet/state.rs
+++ b/sdk-libs/transaction/src/wallet/state.rs
@@ -1,7 +1,7 @@
 use std::collections::{hash_map::Entry, BTreeSet, HashMap, HashSet};
 
 use solana_address::Address;
-use zolana_keypair::{shielded::ShieldedAddress, P256Pubkey};
+use zolana_keypair::{shielded::ShieldedAddress, viewing_key::ViewTag, P256Pubkey};
 
 use crate::{
     error::TransactionError, instructions::transact::OutputContext, utxo::Utxo, AssetRegistry,
@@ -128,19 +128,29 @@ pub struct SyncReport {
     pub unknown_asset_ids: BTreeSet<u64>,
 }
 
-/// One of the indexer streams the wallet resumes.
+/// One key on one of the indexer streams the wallet resumes.
 ///
-/// A selector rather than a key wrapper: the inner maps stay
-/// `HashMap<[u8; 32], Vec<u8>>`, which is what the generic resume-point helpers
-/// already take, so nothing downstream needs to know a stream exists.
+/// The variant carries what its stream indexes by, so the map is flat and a tag
+/// cannot be mistaken for a nullifier -- `ViewTag` is itself `[u8; 32]`, so
+/// keying by the bare value would make the two interchangeable.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum CursorStream {
     /// Shielded transactions matched by output view tag.
-    Tags,
+    Tags(ViewTag),
     /// Shielded transactions matched by spent nullifier.
-    Nullifiers,
+    Nullifiers([u8; 32]),
     /// Encrypted UTXOs, which proofless deposits are read from.
-    Proofless,
+    Proofless(ViewTag),
+}
+
+impl CursorStream {
+    /// The tag or nullifier itself, for building the query its stream is read
+    /// with.
+    pub fn value(self) -> [u8; 32] {
+        match self {
+            Self::Tags(value) | Self::Nullifiers(value) | Self::Proofless(value) => value,
+        }
+    }
 }
 
 pub struct Wallet {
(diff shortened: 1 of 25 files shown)


sdk-libs/transaction/src/wallet/state.rs
```diff
@@ -145,8 +145,18 @@ pub struct Wallet {
...
+    /// For each unspent nullifier, the position through which no spend of it
+    /// exists.
+    ///
+    /// A separate stream from [`Self::sync_cursors`]: reaching the tip of one
+    /// says nothing about the other. Entries are dropped once the nullifier is
+    /// spent.
+    pub nullifier_cursors: HashMap<[u8; 32], Vec<u8>>,
+    /// The same for the encrypted-utxo stream, which proofless deposits are read
+    /// from. Separate for the same reason.
+    pub proofless_cursors: HashMap<ViewTag, Vec<u8>>,
```
ananas-block: why do we have 3 hash maps? 2 would probably be enough or 1 with an enum

sergeytimoshin: refactored

sdk-libs/transaction/src/wallet/state.rs
```diff
@@ -144,9 +159,16 @@ pub struct Wallet {
...
+    /// Sync watermarks, one map per stream.
+    ///
+    /// For each key the stream indexes by -- a view tag, or a nullifier for
+    /// [`CursorStream::Nullifiers`] -- the indexer position through which
+    /// everything matching has already been seen.
+    ///
+    /// The streams are independent: reaching the tip of one says nothing about
+    /// the others, so a shared map would conflate watermarks that advance
+    /// separately. Nullifier entries are dropped once the nullifier is spent.
+    pub cursors: HashMap<CursorStream, HashMap<[u8; 32], Vec<u8>>>,
```
ananas-block: why do we use a hashmap for max 3 items...
We can avoid the nesting with the enum see enum comment.

sdk-libs/transaction/src/wallet/state.rs
```diff
@@ -128,6 +128,21 @@ pub struct SyncReport {
...
+/// `HashMap<[u8; 32], Vec<u8>>`, which is what the generic resume-point helpers
+/// already take, so nothing downstream needs to know a stream exists.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum CursorStream {
+    /// Shielded transactions matched by output view tag.
+    Tags,
+    /// Shielded transactions matched by spent nullifier.
+    Nullifiers,
+    /// Encrypted UTXOs, which proofless deposits are read from.
+    Proofless,
```
ananas-block:
```suggestion
    Tags(ViewTag),
    /// Shielded transactions matched by spent nullifier.
    Nullifiers([u8;32]),
    /// Encrypted UTXOs, which proofless deposits are read from.
    Proofless(ViewTag),
```

ananas-block (approved):

PR #222, opened 2026-08-13
prover: answer a status poll with two key reads instead of four queue scans

sergeytimoshin:
Replaces #210, rebased onto `main`. That PR was stacked on `prover/queue-wait-metric` and GitHub's stack feature refused to retarget it, so it is closed in favour of this one; same commits, no conflicts, `go build ./...` and `go vet ./...` clean.

It also carries #202, which is the first of the two commits here. #202 merged into `prover/queue-wait-metric` at 16:32, after #196 had already squashed that branch's content to `main` at 16:25, so its work never reached `main` and would otherwise be lost.

`checkJobExistsDetailed` runs on every `/prove/status` request — the hottest path in the server, since a client polls it from submission until its proof lands — and answered by walking whole Redis lists with `LRANGE key 0 -1`: the job's own queue, then the processing queue, then `zk_failed_queue`, then `zk_results_queue`, unmarshalling every element of each.

A 220-worker load test made it visible. In one ten-minute window:

| | |
|---|---|
| status polls | 34,154 (~57/s) |
| `zk_results_queue` length | 8,237 |
| implied element reads | ~470k/s from a single-threaded Redis |
| Redis `i/o timeout` | 221 |
| `context deadline exceeded` | 81 |

Dispatch stalls behind the same Redis, health checks fail, ECS replaces the tasks, and the jobs in flight on them are orphaned. Throughput held ~24.5 tps for four consecutive 30s windows, then went to 0.00 and stayed there.

This is the fourth instance of the same shape — a linear Redis list scan on a per-job path — after #202 and its two parents. It mattered most, because it is per poll rather than per job.

The scans were there because job metadata was written once at submission with `status: "queued"` and never updated, so `queued` and `processing` could only be told apart by looking in both queues. Failures were worse: the metadata was deleted on failure, leaving `zk_failed_queue` as the only record, so every poll for an unfinished job scanned that too.

The workers now keep the metadata current:

- `MarkJobProcessing` when a job enters the processing queue
- `MarkJobFailed` on every terminal path — proof error, panic, expiry, replayed cached failure, and the stuck-job reaper — carrying the failure details that used to be read back out of the queue entry

That is one `SET` per job against one scan per poll, and polls outnumber jobs by orders of magnitude. The status path is now `GetResult`, already O(1), plus one metadata `GET`.

The reaper case bites if missed: it fails jobs whose metadata may already have expired, so `MarkJobFailed` recreates the record rather than skipping it. Without that, a reaped job is indistinguishable from an unknown one and the client waits out its entire deadline — which is how 197 workers ended up parked for 35 minutes against an idle prover.

`findJobInQueue` and `extractResultFromPayload` are removed along with their last callers.

`TestFailedJobStatusHTTPEndpoint` built its fixture by writing straight into `zk_failed_queue`, which only worked because of the scan; it now fails the job the way the workers do. Five new tests cover the transitions, that picking a job up preserves its submission expiry rather than renewing it, and that a reaped job still leaves a terminal record.

`go build`, `go vet`, `gofmt` and the full `go test ./...` clean.

This removes the collapse. It does not raise the ceiling: `NullifierTreeUpdateFailed` (7002) was 3,145 of 3,474 failures in the same run and is what caps throughput at ~24 tps. That is separate.

### DIFF
--- a/prover/server/server/queue.go
+++ b/prover/server/server/queue.go
@@ -164,6 +164,77 @@ func (rq *RedisQueue) DeleteJobMeta(jobID string) error {
 	return rq.Client.Del(rq.Ctx, key).Err()
 }
 
+// MarkJobProcessing records that a worker has picked the job up.
+//
+// The metadata used to be written once at submission and never updated, so
+// "queued" or "processing" could only be told apart by walking both queues.
+// The status endpoint did exactly that, on every poll: see the comment on
+// checkJobExistsDetailed. One SET per job replaces a list scan per poll, and
+// polls outnumber jobs by orders of magnitude.
+//
+// KeepTTL preserves the submission-time expiry, so a job that is never
+// completed still ages out on its original schedule.
+func (rq *RedisQueue) MarkJobProcessing(jobID string) error {
+	return rq.updateJobMeta(jobID, func(meta map[string]interface{}) {
+		meta["status"] = "processing"
+		meta["started_at"] = time.Now()
+	}, redis.KeepTTL)
+}
+
+// MarkJobFailed records a terminal failure and the details a caller needs.
+//
+// The failure details live in the metadata rather than only in
+// zk_failed_queue, because reading them back out of that queue means scanning
+// it. The TTL is refreshed to a full hour here: unlike a success, which is
+// answered by the O(1) result lookup, this record is the only thing that can
+// tell a polling client to stop waiting.
+func (rq *RedisQueue) MarkJobFailed(jobID string, details map[string]interface{}) error {
+	return rq.updateJobMeta(jobID, func(meta map[string]interface{}) {
+		meta["status"] = "failed"
+		meta["failure"] = details
+	}, time.Hour)
+}
+
+// updateJobMeta applies mutate to a job's metadata and writes it back.
+//
+// A missing key is recreated rather than skipped. A job can be failed by the
+// stuck-job reaper after its metadata has expired, and a client polling such a
+// job must still be told to stop rather than being left to time out.
+func (rq *RedisQueue) updateJobMeta(
+	jobID string,
+	mutate func(map[string]interface{}),
+	ttl time.Duration,
+) error {
+	key := fmt.Sprintf("zk_job_meta_%s", jobID)
+	meta := map[string]interface{}{}
+	switch existing, err := rq.Client.Get(rq.Ctx, key).Result(); {
+	case err == redis.Nil:
+		// Recreated below with whatever mutate sets.
+	case err != nil:
+		return fmt.Errorf("failed to read job meta: %w", err)
+	default:
+		if err := json.Unmarshal([]byte(existing), &meta); err != nil {
+			return fmt.Errorf("failed to unmarshal job meta: %w", err)
+		}
+	}
+
+	mutate(meta)
+
+	data, err := json.Marshal(meta)
+	if err != nil {
+		return fmt.Errorf("failed to marshal job meta: %w", err)
+	}
+	// A recreated key cannot inherit a TTL that no longer exists, so it gets a
+	// fresh hour rather than KeepTTL's "no expiry at all".
+	if ttl == redis.KeepTTL && rq.Client.TTL(rq.Ctx, key).Val() < 0 {
+		ttl = time.Hour
+	}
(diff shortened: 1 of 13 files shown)


prover/server/server/queue.go
```diff
@@ -155,6 +164,77 @@ func (rq *RedisQueue) DeleteJobMeta(jobID string) error {
...
+
+	mutate(meta)
+
+	data, err := json.Marshal(meta)
+	if err != nil {
+		return fmt.Errorf("failed to marshal job meta: %w", err)
+	}
+	// A recreated key cannot inherit a TTL that no longer exists, so it gets a
+	// fresh hour rather than KeepTTL's "no expiry at all".
+	if ttl == redis.KeepTTL && rq.Client.TTL(rq.Ctx, key).Val() < 0 {
```
ananas-block: `updateJobMeta` can create a metadata key with no expiry: the `TTL` read and the `SET ... KEEPTTL` are two commands. If the key expires between them, `SET` with `KEEPTTL` recreates it without an expiry — a permanent `zk_job_meta_*` key whose status stays `"processing"`, so polls for that job id report "processing" instead of unknown from then on. The window is milliseconds wide, but the failure is permanent. Cheap fix: after the `SET`, if `TTL(key).Val() < 0`, apply `EXPIRE key 1h`.

prover/server/server/queue.go
```diff
@@ -155,6 +164,77 @@ func (rq *RedisQueue) DeleteJobMeta(jobID string) error {
...
+		meta["failure"] = details
+	}, time.Hour)
+}
+
+// updateJobMeta applies mutate to a job's metadata and writes it back.
+//
+// A missing key is recreated rather than skipped. A job can be failed by the
+// stuck-job reaper after its metadata has expired, and a client polling such a
+// job must still be told to stop rather than being left to time out.
+func (rq *RedisQueue) updateJobMeta(
```
ananas-block, line 232: The read-modify-write here is not atomic: GET → mutate → SET with no `WATCH` or Lua. Concurrent writers are last-write-wins; the realistic collision is the worker's failure path against the reaper, and both write `failed`, so at worst one writer's `failure` details are dropped. Acceptable, but this file explains its other tradeoffs in comments and this one deserves a sentence — or the same Lua script that fixes the TTL race below.

prover/server/server/server.go
```diff
@@ -270,155 +280,62 @@ func getStatusMessage(status string) string {
...
-		status := "queued"
-		if metaStatus, ok := jobMeta["status"].(string); ok {
-			status = metaStatus
-		}
-		return true, status, map[string]interface{}{
-			"circuit_type": jobMeta["circuit_type"],
-			"submitted_at": jobMeta["submitted_at"],
-			"from_meta":    true,
-		}
+	if metaErr != nil || jobMeta == nil {
```
ananas-block: A Redis error on the meta lookup reads as "job unknown": `metaErr != nil || jobMeta == nil` folds a transient Redis failure into "never existed", so during a Redis blip clients are told their live job does not exist. Not a regression — the old fallback scans failed the same way — but with the path reduced to two reads, returning a 5xx on `metaErr != nil` so the client retries is a two-line improvement.

prover/server/status_lookup_test.go
```diff
@@ -0,0 +1,156 @@
...
+// carry a status that is actually current. These tests pin that: every
+// transition a job can make must be visible without looking in a queue.
+//
+// The scans they replace were not a micro-optimisation. checkJobExistsDetailed
+// walked up to four whole Redis lists per poll, and a client polls from
+// submission until its proof lands, so the cost scaled with queue length times
+// poll rate. With 8237 entries in zk_results_queue it saturated Redis, and the
+// prover tasks were killed by failing health checks.
+
+func TestJobStatusStartsQueued(t *testing.T) {
```
ananas-block, line 20: Teardown differs from the sibling suites: `dedup_cache_test.go` and `processing_queue_test.go` use `defer teardownRedisQueue(t, rq)`; this file does not. Harmless with miniredis, which cleans up per test, but with `TEST_REDIS_URL` pointing at a shared Redis the suites behave differently. Pick one style.

ananas-block (approved):

PR #223, opened 2026-08-13
indexer: sort the tag query by the table it filters

sergeytimoshin:
`get_encrypted_utxos_by_tags` filtered on `rings_outputs.view_tag` but ordered
by `rings_transactions.*`. No index spans a join, so the planner walked the
transactions table and discarded most of what it read.

## Changes

- Denormalize `signature` and `event_index` onto `rings_outputs`, with a
  backfill and `idx_rings_outputs_view_tag_order` on
  `(view_tag, slot, signature, event_index, output_index)`. Same trade
  `rings_outputs.slot` already makes.
- Point the ORDER BY and the cursor at `po.*`, so filter and sort come from one
  index.
- Rewrite the cursor predicate as a row comparison `(a,b,c,d) > (w,x,y,z)`.
  Postgres can start an index scan at that and cannot at the equivalent OR
  chain, where each page costs more than the last.

## Data

At 64 tags — what the client sends (`DEFAULT_TAG_QUERY_CHUNK`):

| | execution | buffers |
| --- | --- | --- |
| main | 0.728ms | 1191 |
| this branch | 0.655ms | 431 |

### DIFF
--- a/services/photon/src/api/method/rings/common.rs
+++ b/services/photon/src/api/method/rings/common.rs
@@ -283,35 +283,53 @@ pub(super) fn bind_u64_as_i64(
     Ok(bind_sql_value(params, backend, value))
 }
 
-/// Builds the `(slot, signature, event_index)` "after this position" predicate.
+/// Builds the "strictly after this position" predicate as a row comparison.
 ///
 /// `alias` selects which table the predicate reads from, and it must be the
-/// same table the query's ORDER BY uses. Both columns exist on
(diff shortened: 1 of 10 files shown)


ananas-block:
we should enable ordering of both
new to old // better for fetching confidential state since we have a deterministic confidential view tag we 
old to new // better for syncing anonymous state since we fetch view tags starting from index 0
I think currently we default to option 2.
Feel free to make it configurable in this pr or a different one.

ananas-block (approved):

PR #225, opened 2026-08-14
sdk: remember where the proofless scan stopped

sergeytimoshin:
Cherry-picked from #207 (`2a04cc43`), which bundles this with bounded proof
polling, typed indexer errors, and new load-test tooling. Those are unrelated to
sync and shouldn't wait behind it -- and bundling them makes the measurement
unattributable, which is how the branch this stacks on went wrong the first time.

**Stacked on #220.** Review that first; this diff is against it.

## The change

`sync_wallet` kept per-tag cursors for the transaction stream, so each sync
resumed where the last finished. The encrypted-utxo stream, which proofless
deposits are read from, had no such memory: its `scanned_to_tip` set was created
fresh on every call, so every sync paged the entire history for every tag and
discarded all but the deposits client-side.

Measured on devnet in the original commit, 8 workers, wallets carrying ~10k
transfers:

```
sync_wallet_total          2650ms
  fetch_round               970ms
  fetch_proofless_deposits  909ms   <- to keep 2.5 rows
  collect_sort_and_decrypt  341ms
```

My own bench on #220 says the same thing from the other side: with the three
round queries running concurrently as of #200, and the nullifier scan fixed,
`fetch_proofless_deposits` at ~794ms *is* the round's critical path. It is the
one remaining query worth attacking.

## What stacking bought

The original commit closed with a known gap:

> Not yet covered by a unit test: MockIndexer always returns `next_cursor: None`,
> so pinning this needs a mock that advances, and its 18 construction sites all
> take the field list positionally. Worth doing before merge.

#220 built that mock. `TaggedIndexer` already pages and advances, so the missing
test is now cheap: this adds `utxo_entries`/`utxo_calls` to it and
`a_second_sync_resumes_the_proofless_stream`, which asserts the second sync
carries the position the first reached rather than starting over. On the code
before this change the test does not compile -- `fetch_proofless_deposits` takes
no cursors -- so it is a real guard.

The conflict resolution also deleted code rather than adding it: the original
wrote its own resume-point bucketing, twice. #220 extracted that into
`group_by_resume_point`, so both proofless paths now call the shared helper. The
sync path loses 10 lines, the async path another 10, and the rule that a chunk
carries one cursor is stated once for all three streams.

## Scope

What this test pins is what the client *sends* -- that a second sync resumes.
Whether photon reports that position the same way is the local stack's job; CI's
e2e suite runs the real indexer against the real SDK, and that is what caught the
schema break on #220.

### DIFF
--- a/sdk-libs/transaction/src/wallet/state.rs
+++ b/sdk-libs/transaction/src/wallet/state.rs
@@ -157,6 +157,18 @@ pub struct Wallet {
     /// Entries are dropped once the nullifier is spent, since the question is
     /// then answered for good.
     pub nullifier_cursors: HashMap<[u8; 32], Vec<u8>>,
+    /// The same watermarks for the encrypted-utxo stream, which proofless
+    /// deposits are read from.
+    ///
+    /// Separate again for the same reason: reaching the tip of the transaction
+    /// stream says nothing about where the encrypted-utxo stream has been read
+    /// to, and sharing one cursor would skip rows in whichever is behind.
+    ///
+    /// This existed only for the duration of a single sync before, so every sync
+    /// re-read the whole encrypted-utxo history for every tag and discarded all
+    /// but the deposits. Measured on devnet: 909ms per sync to keep 2.5 rows,
+    /// about a third of the sync phase, and growing with history.
+    pub proofless_cursors: HashMap<ViewTag, Vec<u8>>,
 }
 
 impl Wallet {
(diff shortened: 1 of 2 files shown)


sergeytimoshin:
Collapsed into #220 rather than kept as a stack. A stacked PR only pays off if its base moves; #220 is blocked on review with nothing in the queue moving, so this would have been unreviewable on its own while inheriting every rebase #220 needs.

Commit `291f236c` is unchanged, including the resume test the original commit said it could not write — #220 had since built the mock that made it possible.

PR #228, opened 2026-08-17
feat: let a hardware-resident keypair back a wallet authority

sergeytimoshin:
**Motivation:**
1. `ShieldedKeypairTrait` has no consumer, so implementing it buys a backend nothing. The wallet takes `WalletAuthority`, and both impls are hardwired to `ShieldedKeypair`.
    - a backend can implement `SyncWalletAuthority` by hand, but the four bodies it needs are private free functions here, so it either duplicates protocol-critical encryption glue or holds a throwaway local `ShieldedKeypair` to borrow them — which detaches the published identity from the device key
2. `RoleExpansion` is `pub(crate)`, so an out-of-crate backend cannot expand a device-produced seed at all without reimplementing the HKDF schedule. A divergent expansion yields a valid-looking *different* identity: signatures verify, addresses look right, nothing catches it.
3. The 12-method `ViewingKeyTrait` forwarding block is already written out 3x at 80 lines each, with nothing keeping the copies in step.

**Changes:**
1. add `derivation::expand_roles(seed, curve)`, one public entry for both role secrets
    - validates seed width per rail (64 ed25519, 32 P-256); HKDF-Extract accepts any length, so an unchecked truncated seed would expand into that same wrong-but-valid identity
    - `KeypairError` becomes `#[non_exhaustive]`: out-of-crate backends match on it, so a future failure mode should not be a breaking change for them
    - `RoleExpansion` stays private: publishing it means publishing `Rail` alongside `Curve`, two overlapping public enums for one concept, and no caller wants the type, only the secrets
    - the in-tree KMS tests keep their hand-rolled expansions on purpose. They cross-check the schedule against the published `INFO_*` constants; routing them through this function would make the assertion tautological
2. make `KeypairWalletAuthority` generic over the keypair, `K = ShieldedKeypair`
    - the four helper bodies only ever needed `ViewingKeyTrait` + `ShieldedKeypairTrait`
    - `K = ShieldedKeypair` because a software keypair is the ordinary case, the way `HashMap` defaults its hasher. It also happens to leave existing call sites untouched, but that is not the reason — the type's old name was removed outright, with no alias
    - `with_viewing_keys` takes current + historical keys, which the old constructor could not express (a `ShieldedKeypair` holds one)
    - fallible, with its own `AuthorityViewingKeyMismatch`: encryption keys off the keypair while a scan keys off the vector, so a mismatched pair would encrypt to one key and scan with another. `new` stays infallible, it reads the key off the keypair. Distinct from `MissingCurrentViewingKey`, which means a scan was handed a stale snapshot
3. add `forward_viewing_key_trait!`, applied at all 4 forwarding sites
    - not a blanket impl over an accessor trait: coherence would stop a downstream crate implementing `ViewingKeyTrait` by hand, which is exactly what an HSM-resident viewing key needs
    - `impl ViewingKeyTrait for ViewingKey` stays hand-written, it is the base case the macro forwards to
4. rename `LocalWalletAuthority` -> `KeypairWalletAuthority`, Rust and TypeScript
    - generalizing the type made the old name false: it now accepts a keypair that is explicitly not local
    - no alias, the crate is unpublished so nothing external holds the old name
    - renamed in TS too so the two SDKs do not drift; `authority-v1.json`'s `rustSymbol` moves with it and its manifest digest is recomputed, other 4 fixtures unchanged
    - separate commit, droppable if you would rather keep the name

**Testing:**
1. `a_remote_keypair_backend_is_a_wallet_authority` drives a double that is deliberately not a `ShieldedKeypair`; signing counted through the trait, identity/signature/`tx_viewing_pk` pinned to the software authority. Equality against the software path is the only assertion that catches a wrong-but-well-formed identity.
    - also covers the mismatched-viewing-key rejection, and that the backend's derivation-seed guard still refuses through the authority rather than only on a direct trait call
2. `expand_roles` pinned against `from_keypair` on both rails, comparing secrets rather than pubkeys; plus wrong-width and PDA cases
3. `cargo test -p zolana-keypair -p zolana-transaction -p zolana-wallet`, `clippy --workspace --all-targets -- -D warnings`, `fmt --all --check`, `doc --no-deps`
4. `npm run check:ts`: oxfmt, oxlint, `tsc --noEmit`, 243 vitest tests, build, example project

**Notes:**
1. TS fixtures are sha256-verified against their manifest and `cargo` passes regardless. `npm run check:ts` is needed for anything touching `sdk-libs/ts`, including Rust-looking changes, since fixtures name Rust symbols.
2. a Turnkey backend built on this stacks in `keypair/turnkey-backend`; it is what surfaced a missing `ViewingKeyTrait` impl on one of its own rails.

### DIFF
--- a/cli/src/wallet_cli/material.rs
+++ b/cli/src/wallet_cli/material.rs
@@ -22,7 +22,7 @@ use zolana_transaction::{
 };
 use zolana_wallet::{
     AnonymousRecipientSlot, ApprovalRequest, EncryptedSplit, EncryptedTransfer,
-    LocalWalletAuthority, P256Signature, SyncWalletAuthority,
+    KeypairWalletAuthority, P256Signature, SyncWalletAuthority,
 };
 
 use super::{resolve::ResolvedSyncOptions, util::parse_hex_array};
@@ -101,7 +101,7 @@ impl SyncWalletAuthority for WalletMaterial {
         assets: &AssetRegistry,
     ) -> std::result::Result<EncryptedTransfer, TransactionError> {
         SyncWalletAuthority::encrypt_confidential_transfer(
-            &LocalWalletAuthority::new(self.solana_pubkey(), &self.keypair),
+            &KeypairWalletAuthority::new(self.solana_pubkey(), &self.keypair),
             first_nullifier,
             outputs,
             assets,
@@ -116,7 +116,7 @@ impl SyncWalletAuthority for WalletMaterial {
         recipients: &[AnonymousRecipientSlot],
     ) -> std::result::Result<EncryptedTransfer, TransactionError> {
         SyncWalletAuthority::encrypt_anonymous_transfer(
-            &LocalWalletAuthority::new(self.solana_pubkey(), &self.keypair),
+            &KeypairWalletAuthority::new(self.solana_pubkey(), &self.keypair),
             first_nullifier,
             sender_view_tag,
             sender,
@@ -131,7 +131,7 @@ impl SyncWalletAuthority for WalletMaterial {
         bundle: &SplitBundlePlaintext,
     ) -> std::result::Result<EncryptedSplit, TransactionError> {
         SyncWalletAuthority::encrypt_split(
-            &LocalWalletAuthority::new(self.solana_pubkey(), &self.keypair),
+            &KeypairWalletAuthority::new(self.solana_pubkey(), &self.keypair),
             first_nullifier,
             view_tag,
             bundle,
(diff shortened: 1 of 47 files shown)


sergeytimoshin:
Superseded by `sdk/remote-wallet-authority`, which carries this work forward rather than dropping it.

Every change here is in that branch: `derivation.rs`, `error.rs`, `pda.rs` and `traits/view_key.rs` are byte-identical, and the rest (`expand_roles` with per-rail width validation, `KeypairError` `#[non_exhaustive]`, `forward_viewing_key_trait!`, `KeypairWalletAuthority<K = ShieldedKeypair>` with `with_viewing_keys` and `AuthorityViewingKeyMismatch`, the `LocalWalletAuthority` rename across Rust and TypeScript, and the `authority-v1.json` fixture digest) is in its `feat: support remote wallet authorities` commit. That branch then builds on it: a client-owned Ed25519 authority, local verification of prover responses, and wallet-sync fixes.

`keypair/turnkey-backend` has been rebased onto the new branch and re-pushed, so the Turnkey backend is not stranded. It builds and its 35 tests pass against the new base.

Closing to keep one review instead of two. Nothing to re-do here.

PR #231, opened 2026-08-17
chore(ts-sdk): align key derivation with Rust

ananas-block:
## Summary

The TS package (`sdk-libs/ts`) had drifted from `sdk-libs/keypair` in both derivation logic and naming. Rust is the source of truth; this PR makes the TS derivation byte-identical and pins cross-language parity with shared test vectors so it cannot drift silently again.

### Additional Changes
- remove Viewingkey::from seed
- add ts test from seed


tilo-14 (approved):
lgtm

PR #232, opened 2026-08-17
custom-ring: add a minimal custom ring with auditor verifiable encryption

ananas-block:
Adds `custom-ring-tests/`, a minimal custom ring example mirroring `sdk-tests/zk-program-swap`. Its single feature is verifiable encryption to an auditor: the circuit proves the per-transaction viewing secret key is correctly encrypted to the auditor P256 key held in the ring's config account, and that the published `tx_viewing_pk` is that secret's public key. Every user ciphertext in an SPP transaction is already HPKE'd under the transaction viewing key, so an auditor who recovers it decrypts everything the recipients see and the circuit needs no per-output witnesses. User encryption itself is not checked.

Scope is deliberately minimal: no smart accounts, no user accounts, no key rotation or viewing-key lifecycle, confidential only, and eddsa signers only (`CircuitId::RingEddsa` -> SPP `RING_TRANSACT`; `RingP256` is rejected). SPP needs no changes, since the ring hooks it already exposes are sufficient.

## Crates

- **program**: the config account, SPP registration, a deposit forwarder, and transact, which verifies the audit proof and then CPIs SPP with the `ring_auth` PDA flipped to a signer
- **prover**: the gnark circuit (327883 constraints, one BSB22 commitment over private wires) plus its cgo/bindgen bridge and the trusted-setup binary
- **sdk**: instruction builders, the host mirror of the circuit's crypto, and the two-step proof-input API
- **client**: the auditor side, which recovers the transaction viewing key and returns typed decrypted outputs
- **test**: localnet bring-up plus the end-to-end auditor lifecycle

## Design notes

The auditor ciphertext rides in the SPP transact messages, so it is bound both by the ring proof and by SPP's own `external_data_hash`, and Photon indexes it with no changes. The recovered key opens Confidential transact output slots only; ring deposits stay public and are read rather than decrypted.

Cross-language agreement is pinned by tests rather than by inspection: the public input hash is fixed to a value computed in Go with the same Poseidon the in-circuit gadget links, and both that fixture witness and the witness the sdk produces verify against the committed verifying key and the on-disk `vk.bin`, so key drift fails the suite.

## Keys and CI

Proving keys are gitignored and pinned by `custom-ring-keys.CHECKSUM`. The `custom-ring-keys-v1` release is not published yet, so the new CI job cannot pass on a clean checkout until it exists.

## Testing

95 tests pass locally: 44 program, 34 sdk, 14 client, 2 localnet.


services/ring-rpc/src/audit.rs
```diff
@@ -0,0 +1,333 @@
+//! Reading ring transactions by the auditor tag and opening them.
```
ananas-block: what's the auditor tag?

services/ring-rpc/src/page.rs
ananas-block, line 1: why does the server have a page?

services/ring-rpc/src/audit.rs
Atamanov: The view tag the auditor scans for.   X coordinate of the pk ( auditor_view_tag(pk) -> [u8;32] { pk.x() }),

tilo-14 (approved):

PR #235, opened 2026-08-18
test(spp): move the Groth16 suites into their own package

sergeytimoshin:
Split from #184, which bundled this with the test-hygiene checker rewrite. This
half stands alone and fixes a hole worth landing on its own.

## The hole

Six test targets in `shielded-pool-tests` carry `required-features = ["proofs"]`:
`merge_functional`, `transact_functional`, `transact_withdrawal`,
`mixed_interface_transfers`, `transact_settlement`, and `bench_cu`.

CI runs `just check-all` (`cargo check --workspace --all-targets`) and
`just clippy` (`cargo clippy --workspace --all-targets -- -D warnings`). Neither
passes `--features proofs`, and cargo skips a target whose `required-features`
are not enabled even under `--all-targets`.

So those six compile in neither `check` nor `clippy` on any PR. They are built
only by the proofs tier, which needs a prover, proving keys, and a built CLI --
so a break there stays invisible until someone runs it. That is how the CU bench
stopped compiling at #172 without anything going red.

## The change

The feature flag bought hermeticity for the fast tier: a plain `cargo test -p
shielded-pool-tests` must not try to spawn a prover. A package boundary buys the
same thing -- another package is simply unreachable from `-p
shielded-pool-tests` -- while keeping the files in every `--all-targets`
compile. Strictly better on both axes.

`shielded-pool-proofs` uses a flat `tests/` layout so cargo auto-discovers each
file as a target: no `[[test]]` blocks to keep in sync, and no way to add a
suite here that never runs. File names match the previous target names, so
recipe and CI selectors are unchanged.

`bench_cu` stays in `shielded-pool-tests` and simply loses its gate, so it
compiles on every PR again.

## Verification

`cargo metadata` after the split:

| package | test targets | gated |
| --- | --- | --- |
| `shielded-pool-proofs` | 5 | 0 |
| `shielded-pool-tests` | 34 | 4 (`localnet_*`, a separate tier) |

- `cargo check --workspace --all-targets` clean -- and now compiles all five
- `cargo clippy -p shielded-pool-proofs --all-targets -- -D warnings` clean;
  these suites had never been linted, so this confirms no accumulated debt
- `cargo fmt --all --check` clean
- The five suites are pure renames (git reports 100%/99% similarity) and were
  removed from the old crate, not copied

## Not addressed

The four `localnet_*` targets are still gated behind `localnet` and have the
same invisibility. The same package-boundary treatment would fix them; left for
a follow-up rather than widening this change.


sergeytimoshin:
Superseded by #238, which does this without a new package.

The premise here was that a slow tier has to be a package, because the
`required-features` gate that isolates it also hides it from `cargo check
--workspace --all-targets`. The first half is wrong: declaring a `[[test]]` block
for an auto-discovered file **overrides** the inferred target rather than adding a
second, so auto-discovery and `required-features` coexist. Verified against
`cargo metadata` -- 30 targets, one `admin_functional` entry, carrying its
`required-features`.

And the visibility hole it was written for needs two lines, not a restructure:
naming both features in `just check-all` and `just clippy`. Cargo skips a gated
target even under `--all-targets`, so those recipes were the whole problem.

So #238 flattens every suite for cargo to discover, keeps a block only for the ten
gated ones, and adds those two lines. No new packages, and none of the 47
dependency lines this PR and #236 hand-wrote between them -- which had already
produced one missing-dep bug and one CI failure.

The F-07 gate fix (`d0d74c4f`) is carried into #238 at the file's new path: the
local `program_test()` returning `None` on a missing `.so` meant the suite skipped
itself silently, and the fast tier now reports 0 skipped where it reported 1.

Branch kept if anything needs recovering.

PR #236, opened 2026-08-18
test: make every tier a package, and delete the hygiene checks that made moot

sergeytimoshin:
Stacked on #235 -- review that first; this diff is against it, not against main.

Replaces the checker half of #184 (closed). #184 proposed 1243 lines of xtask to
police test-suite structure. This deletes the problem instead, using cargo's own
auto-discovery, and ends up removing code rather than adding it.

## Every tier is a package

#235 made the Groth16 suites a package because a `required-features` gate hides a
target from `cargo check --workspace --all-targets`. The four localnet suites
were in the same hole, so they get the same treatment: `shielded-pool-localnet`.

The `localnet` feature stays, because it had a second job -- it selects
`support::forester` and zolana-client's indexer/RPC surfaces. It just no longer
gates a target. The new package turns it on through its dependency, so the suites
compile in every `--all-targets` build while a plain `cargo test -p
shielded-pool-tests` still pays for neither.

`bench/cu.rs` loses `#![cfg(not(feature = "localnet"))]`. Feature unification now
enables `localnet` across a workspace build, and that attribute would blank the
file out -- reintroducing exactly the invisibility being removed here. It
compiles with the feature on, so the guard was defensive, not load-bearing.

## Suites are auto-discovered

The remaining 30 suites move to top-level `tests/*.rs`. Cargo discovers each as a
binary named after the file, so `autotests = false` and all 30 `[[test]]` blocks
go: the manifest drops from 215 lines to 86, and a suite added under `tests/`
cannot fail to become a target.

Target names are unchanged -- `cargo metadata` reports the same 30 before and
after -- so no justfile or CI selector moves.

## The script shrinks to what still has a job

320 lines to 186. Removed:

- Declared `[[test]]` paths existing on disk: vacuous, there are none.
- Every leaf being a declared target or an included submodule: narrowed to the
  case that survives. A file in a *subdirectory* of `tests/` is not
  auto-discovered, so it runs only if a target pulls it in as a module -- today
  `shielded-pool-localnet/tests/localnet_photon/`. The check now spans all of
  program-tests instead of shielded-pool alone, and handles `tests/common/mod.rs`,
  which names its module after the directory rather than the file.
- The obsolete `tests/common` wrapper guard: that migration is long done.
- The README tallies, and the prose of a retired N/A entry: bookkeeping about the
  bookkeeping. A wrong count misleads nobody about whether an invariant is tested.

Kept, because nothing else can do it: that every `Covered by:` citation resolves,
and that no runtime artifact is committed under a test package. The citations are
the substitute for coverage on the surface llvm-cov cannot see -- the program runs
inside the SVM -- so a citation naming a symbol that no longer exists reads as a
green checkmark against an invariant nothing tests.

`cargo nextest list` was considered for that check and does not fit: citations
also name Go functions and non-test Rust functions, and one form is file-scoped,
while nextest enumerates Rust tests only. It would validate a subset and reject
legitimate citations.

For the record, two claims in #184's commit messages do not hold for the script
it replaced: token interpolation is not injectable (tokens are filtered to
`^[a-z_][a-z0-9_]*$|^Test[A-Za-z0-9]+$` first), and `fn name(` matches multi-line
signatures. The one real gap is generics, `fn name<T>(`.

## Verification

- `cargo metadata`: 30 test targets before and after, no name differences
- `cargo nextest run -p shielded-pool-tests`: 262 passed
- `cargo check --workspace --all-targets`, `clippy --workspace --all-targets`,
  `cargo fmt --all --check`: clean
- `cargo machete` clean on the new package (16 copied-over deps pruned)
- Both surviving hygiene checks were run against a planted violation: each
  reports it and exits 1, verified separately from the message, since printing an
  error while exiting 0 would leave CI green
- `tests/deposit/mutation.proptest-regressions` moved with its suite. It holds a
  real recorded failing input, and proptest resolves that path from the source
  file, so leaving it behind would have orphaned the case with every test still
  passing.


sergeytimoshin:
Superseded by #238. The flattening and the script slimming survive there; the two
new packages do not.

This PR added `shielded-pool-localnet` for the same reason #235 added
`shielded-pool-proofs`: the belief that an auto-discovered target cannot carry
`required-features`, so a gated tier has to become a package before the suites can
be flattened. That is false -- declaring a `[[test]]` block for an auto-discovered
file overrides the inferred target rather than adding a second, and the block keeps
its `required-features`. Checked with `cargo metadata`: 30 targets, one
`admin_functional`, gate intact.

#238 therefore flattens all 39 suites into `shielded-pool-tests`, keeps a block for
only the ten gated ones, and closes the check/clippy visibility hole with two
justfile lines instead of a package boundary. Manifest 215 -> 141, script 320 ->
193, no new packages, and none of the hand-written dependency manifests these two
PRs needed -- which had already cost a missing `solana-compute-budget-interface`
and a `cargo-machete` CI failure on `serial_test`.

What is carried across: the flattening itself, the narrowed subdirectory-orphan
check (now spanning all of program-tests, and handling `tests/common/mod.rs`
naming its module after the directory), the `bench/cu.rs` cfg removal, and the
proptest regressions file moving with its suite.

Branch kept if anything needs recovering.

PR #242, opened 2026-08-18
custom-ring: upgrade-authority gate, ring rpc, cargo-generate wizard and pipeline

ananas-block:
Stacked on #232. 

- `create_config` is gated on the program upgrade authority. The CPI forwards through the heap path with a named error.
- `zolana-ring-client` holds the auditor codec and decryption, moved out of the example.
- `services/ring-rpc` holds a ring's auditor key, serves decrypted transactions with sender and recipient over JSON-RPC and as a server-rendered page. Loopback bind, opt-in origins, timeouts, key file permission check.
- `templates/custom-ring` and `just ring-new` generate a ring repository. `custom-ring-cli` deploys, initializes and transacts. `just ring-localnet` boots the services.
- `hello` is a toggleable smoke feature.

Test:

```
just ring-new ~/rings          # wizard, answer the prompts
cd ~/rings/<name>
just localnet                  # validator, photon, prover from this checkout
just pipeline                  # build, deploy, init, ring rpc (stays up), transact
open http://127.0.0.1:8785/    # auditor page, port 8785 plus ZOLANA_PORT_OFFSET
just localnet-stop
```


custom-rings/cli/src/ring_rpc.rs
```diff
@@ -69,30 +115,48 @@ impl RingRpc {
...
+            .duration_since(std::time::UNIX_EPOCH)
+            .map(|elapsed| elapsed.as_secs())
+            .unwrap_or_default();
+        let message = read_attestation(
+            &ring,
+            timestamp,
+            cursor.as_ref().map(|cursor| cursor.0.as_slice()),
+            None,
+        );
+        let request = GetDecryptedTransactionsRequest {
```
ananas-block: why don't we add a constructor for GetDecryptedTransactionsRequest that does the signing and keep 
`request: &GetDecryptedTransactionsRequest,` as input for `decrypted_transactions` ?

Or a builder pattern:
```
GetDecryptedTransactionsRequest::new(ring: Address, cursor: Option<Base64String>).sign(reader: &dyn Signer)?;
```

Atamanov: Implemented that

PR #243, opened 2026-08-19
client: fetch spend proofs concurrently, and stop confirming twice

sergeytimoshin:
Two places where the client waited on itself, found by profiling a transfer against devnet.

## Where a transfer's time goes

`ZOLANA_TIMING` on the deployed stack, p50 over ~110 transfers:

| span | before | owner |
|---|---|---|
| `finish_submission` | 1549ms | — |
| ├ `fetch_spend_proofs` | 774ms | indexer |
| ├ `prove_transfer` | 257ms | prover |
| `send` | 1032ms | RPC |
| `sync_wallet` | 505ms | indexer |
| `confirm` | 207ms | RPC + indexer |

Indexer 42%, RPC 32%, prover 8%. Worth stating plainly because the prover is the part everyone assumes: at 16 vCPU the proof is a quarter of a second, and a 4x hardware bump moved the total by 14%.

## Two fixes

**Spend proofs fetched serially.** The comment above them already said *"Two independent indexer round trips (317ms and 114ms on devnet) that ran back to back. Nothing links them"* — and `fetch_spend_proofs_async` has always used `try_join!`. Only the blocking path was sequential. It now uses `thread::scope`, the same thing its own caller does one level up.

**`send_transaction_with_config` did not send with config.** It called `send_and_confirm_transaction_with_spinner_and_config`, so asking to set a config also bought a confirmation wait. Every caller then paired it with `confirm_private_transaction_sync`, which confirms *again* — one signature, two waits. It now sends; `send_transaction` remains the send-and-confirm one. Nothing outside `client.rs` called it.

`submit_private_transaction_sync` is the pairing that follows, and it reads `send_config` — a field `with_send_transaction_config` has been setting that **nothing ever read**, so a caller who wants preflight can still ask for it.

## Result

Same wallets, same 4 vCPU prover, 180s runs:

| phase | before | after |
|---|---|---|
| send | 1032ms | **133ms** |
| confirm | 207ms | 820ms |
| `fetch_spend_proofs` | 774ms | 684ms |
| **total p50** | **3554ms** | **3229ms** |

The confirmation did not disappear — it moved to `confirm`, where it is paid once. The saving is the duplicate. Being honest about the size: parallelising the proof fetches bought only ~90ms, because the two calls are lopsided rather than equal.

## Test change

The mock indexer now dispatches by request path. Serving responses in position order described a client that fetched sequentially, and stops being true the moment two fetches race — the one test on that path was asserting a request order that is no longer defined, so it now asserts the racing pair as a set.

## Verification

`just check-all`, `just clippy`, `cargo fmt --check`, `cargo machete --skip-target-dir` clean. 55 client tests, 46 CLI tests. The new tests were confirmed to fail when `skip_preflight` is flipped.

### DIFF
--- a/cli/src/wallet_cli/transaction.rs
+++ b/cli/src/wallet_cli/transaction.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use solana_signer::Signer;
-use zolana_client::{SolanaRpc, ZolanaClient};
+use zolana_client::{Rpc, RpcSendTransactionConfig, SolanaRpc, ZolanaClient};
 use zolana_transaction::Address;
 use zolana_wallet::{
     actions::{submit::MergeMaterial, transaction::is_plain_utxo},
@@ -15,6 +15,15 @@ use super::{
     util::{ensure_positive, format_address, parse_address, parse_hex_array, parse_pubkey},
 };
 use crate::args::{MergeOptions, SplitOptions, TransferOptions, UtxosOptions};
+
+/// The CLI confirms separately, so preflight would only buy a simulation
+/// round trip before a send that is about to be confirmed anyway.
+fn send_config() -> RpcSendTransactionConfig {
+    RpcSendTransactionConfig {
+        skip_preflight: true,
+        ..RpcSendTransactionConfig::default()
+    }
+}
 
 pub(crate) fn run_transfer(opts: TransferOptions) -> Result<()> {
     ensure_positive(opts.amount)?;
@@ -46,7 +55,9 @@ pub(crate) fn run_transfer(opts: TransferOptions) -> Result<()> {
         &client,
         &ctx.material.funding,
     )?;
-    let signature = client.submit_private_transaction_sync(&transaction)?;
+    let signature = client
+        .rpc()
+        .send_transaction_with_config(&transaction, send_config())?;
     client.confirm_private_transaction_sync(signature)?;
     let mode = if transfer.recipient.is_public_withdrawal() {
         "withdraw"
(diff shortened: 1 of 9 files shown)


sdk-libs/client/src/client.rs
```diff
@@ -398,6 +398,32 @@ impl<R: Rpc> ZolanaClient<R> {
...
+    /// Pair with [`Self::confirm_private_transaction_sync`], which waits for the
+    /// chain and then the indexer. `Rpc::send_transaction` confirms too, so the
+    /// two together waited for the same signature twice.
+    ///
+    /// Preflight is skipped: simulation re-executes a transaction whose inputs
+    /// are proofs the indexer just served, and a failure it would catch is one
+    /// this client cannot fix by retrying. The cost of being wrong is a landed
+    /// transaction that fails on chain, so callers submitting unvalidated
+    /// transactions should keep using `send_transaction`.
+    pub fn submit_private_transaction_sync(
```
ananas-block: in line with existing solana terminology the following is more intuitive to me

```suggestion
    pub fn send_private_transaction_sync(
```

ananas-block: tbh I don't like the sync suffix very much, why don't we have a async and sync traits that have the same methods with the only difference of being sync or async ?

Should be pretty straightforward to import one trait or the other when using it.

ananas-block: also why is it named private transaction at all  its just sending a solana transaction

cli/src/wallet_cli/transaction.rs
```diff
@@ -46,7 +46,7 @@ pub(crate) fn run_transfer(opts: TransferOptions) -> Result<()> {
         &client,
         &ctx.material.funding,
     )?;
-    let signature = client.rpc().send_transaction(&transaction)?;
+    let signature = client.submit_private_transaction_sync(&transaction)?;
```
ananas-block: I feel like this minor api improvement does not warrant the additions in the client crate

ananas-block:
I don't like that we keep adding public api to the client crate without a plan

ananas-block (approved):

PR #245, opened 2026-08-19
indexer: split the paging cursor from the sync watermark

sergeytimoshin:
Stacked on #243.

`next_cursor` answered two questions with one field, and on a last page they want opposite answers.

- As **"where did I get to"** it must be present — or the caller has no watermark and rescans its whole history every sync. That's what `a_short_page_still_reports_its_last_row` was added to prevent.
- As **"is there another page"** it must be absent — or the caller spends an entire request being told there is nothing.

It was resolved in favour of the watermark, so every sync paid an extra round trip. Measured on devnet with an instrumented client: **40 of 40 transfers made a second request**, worth **170ms of a 339ms fetch**, on both the tag and proofless streams.

## The split

`scanned_through` already existed for the nullifier stream (whose pages are usually empty, so they carry no last row). It now covers the tag and proofless streams too and carries the watermark. `next_cursor` means only that another page exists, which the endpoints establish by fetching **one row past the limit and returning it to nobody**.

**The client needed no change.** `advance` already preferred `scanned_through` for the watermark and already used `next_cursor` to decide whether to loop — the design anticipated this.

The outputs stream needed its own scan position: its cursor carries `output_index`, so the tail of `rings_transactions` doesn't describe where an output scan reached.

## Why it's not the naive version

The obvious fix — drop the cursor on a short page — is the regression the existing test guards, and I wrote it before reading that test. It would have made every wallet whose history fits in one page refetch all of it, every sync. Trading 170ms for that would have been a bad deal.

## Test changes, and one that's a real improvement

Three wallet tests asserted the two-request behaviour and now assert one. Their substantive assertions — that the cursor still advances to the tip, and that a tipped tag is never re-queried — pass unchanged, which is the evidence the watermark survived.

`a_truncated_page_reports_no_scan_position` now inserts a second matching transaction. Both fixture nullifiers belong to one transaction, so a limit of 1 returned a page that filled *exactly* and was never truncated — the old rule couldn't tell that from a real truncation and assumed the worse of the two. Over-fetching distinguishes them, so the watermark is now reported in a case where it was previously withheld.

## Measurements

Locally **492ms p50, unchanged** from 489ms — as expected, since this removes a round trip and local RTT is ~0. The tail did improve: p99 1290ms → 590ms.

The devnet effect follows from RTT and is not yet measured on the deployed stack; the image is built (`cursorsplit-91c4a710d387`) if you want it deployed.

## Verification

`just check-all`, `just clippy`, `cargo fmt --check`, `cargo machete --skip-target-dir` clean. 88 photon, 112 wallet, 263 TS (`npm run check`: format, lint, typecheck, tests, build). openapi regenerated.

### DIFF
--- a/sdk-libs/client/src/indexer.rs
+++ b/sdk-libs/client/src/indexer.rs
@@ -30,7 +30,15 @@ use crate::{
 };
 
 const MERKLE_PROOF_POLL_TIMEOUT: Duration = Duration::from_secs(60);
-const MERKLE_PROOF_POLL_INTERVAL: Duration = Duration::from_millis(500);
+/// First wait after an incomplete answer.
+///
+/// A transfer spends a note the indexer has only just written, so the first
+/// attempt often lands a few milliseconds early and this sleep is on the
+/// critical path of every transfer. A flat 500ms charged the tail's wait to
+/// every caller; starting short and backing off keeps the 60s ceiling for an
+/// indexer that is genuinely behind.
+const MERKLE_PROOF_POLL_START: Duration = Duration::from_millis(25);
+const MERKLE_PROOF_POLL_MAX: Duration = Duration::from_millis(500);
 
 const JSON_RPC_METHOD_NOT_FOUND: i64 = -32601;
 const JSON_RPC_INTERNAL_ERROR: i64 = -32603;
@@ -357,6 +365,7 @@ impl Rpc for ZolanaIndexer {
         let expected = leaves.len();
         let started = Instant::now();
         let mut last_error = None;
+        let mut wait = MERKLE_PROOF_POLL_START;
         loop {
             match single() {
                 Ok(response) if response.proofs.len() >= expected => return Ok(response),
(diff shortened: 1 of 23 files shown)


sdk-libs/client/src/indexer.rs
ananas-block, line 341: is this dead code?

sdk-libs/client/src/rpc.rs
```diff
@@ -59,13 +59,19 @@ pub struct GetEncryptedUtxosByTagsResponse {
     pub context: Context,
     pub matches: Vec<EncryptedUtxoMatch>,
     pub next_cursor: Option<Vec<u8>>,
+    /// Where the indexer's scan reached. The watermark to resume from, which
+    /// `next_cursor` no longer carries on a last page.
+    pub scanned_through: Option<Vec<u8>>,
```
ananas-block, line 70: shouldn't we fix cursor then instead of adding a new equivalent value?

PR #246, opened 2026-08-19
keys: serve the example circuits' proving keys from the object store

sergeytimoshin:
The SPP proving keys were moved to S3 behind CloudFront with hashes pinned in a lockfile the prover embeds. **The four example circuits never followed** — custom-ring, swap, dynamic-swap and timelock-escrow each still ran their own `gh release download`.

That drift has a cost. On #243 the `tests / custom-ring` job went red on:

```
Get "https://release-assets.githubusercontent.com/…auditor_key_encryption_vk.bin":
  read tcp …: connection reset by peer
error: recipe `ensure-custom-ring-keys` failed with exit code 1
```

The key never reached a compile. Four recipes share that failure mode, and all four also needed a GitHub token to fetch parameters that aren't secret.

## Change

| | before | after |
|---|---|---|
| source | `gh release download` ×4 | CloudFront, one shared recipe |
| auth | GitHub token | none |
| override | — | `ZOLANA_PROVING_KEYS_URL` (same var as the SPP keys) |
| integrity | hash checked after a download | hash checked **every run** |
| partial fetch | could be trusted next run | lands on `.part`, moved only when complete |

Keys live at `proving-keys/examples/<tag>/`. **Not** a top-level prefix of its own: the bucket policy grants CloudFront `proving-keys/*` and that policy is managed outside this repo, so a new prefix meant either a hand-edited policy drifting from its source or a 403. I tried the 403 first.

`just publish-example-keys <tag> <base>` is the other half — rotation now uploads instead of cutting a release, and it **refuses a tag whose folder already exists**, since a checkout pinned to that tag would otherwise get different bytes. A rotation writes a new folder and leaves the old one readable.

The GitHub releases are untouched. Nothing reads them now, but they're the way back.

## Verification

All 18 files were downloaded from the releases, checked against the committed manifests, and uploaded only after every hash matched. Then, with every cached key deleted:

- all four recipes fetch over CloudFront and exit 0 — **18 files, each matching its committed hash**
- appending a byte to one is rejected with the mismatch printed
- deleting one refetches it
- a second run re-hashes in 0.3s without refetching
- `publish-example-keys swap-keys-v4 …` refuses, because that folder exists
- `cargo test -p custom-ring-sdk` passes against the fetched keys
- every dependent recipe still resolves (`test-custom-ring`, `bench-swap`, `test-swap-validator`, `test-custom-ring-validator`, `test-escrow-validator`, …)

## Not included

The SPP keys' lockfile is embedded in the prover binary so pk↔vk↔code can't drift. These manifests stay as committed `*.CHECKSUM` files read by the recipe, because the consumer is a test-time `just` recipe rather than a running binary. Worth revisiting if these keys ever get loaded at runtime.

### DIFF
--- a/cli/src/wallet_cli/transaction.rs
+++ b/cli/src/wallet_cli/transaction.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use solana_signer::Signer;
-use zolana_client::{Rpc, SolanaRpc, ZolanaClient};
+use zolana_client::{SolanaRpc, ZolanaClient};
 use zolana_transaction::Address;
 use zolana_wallet::{
     actions::{submit::MergeMaterial, transaction::is_plain_utxo},
@@ -46,7 +46,7 @@ pub(crate) fn run_transfer(opts: TransferOptions) -> Result<()> {
         &client,
         &ctx.material.funding,
     )?;
-    let signature = client.rpc().send_transaction(&transaction)?;
+    let signature = client.submit_private_transaction_sync(&transaction)?;
     client.confirm_private_transaction_sync(signature)?;
     let mode = if transfer.recipient.is_public_withdrawal() {
         "withdraw"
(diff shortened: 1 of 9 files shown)


ananas-block:
hm not sure this is the best solution, it would probably be better if we did a setup with constant randomness so that we always get the same provingkey and don't need to publish it at all. I already did that somewhere not sure on which branch or which repo.

PR #248, opened 2026-08-19
fix: close ECDH point-negation guard bypass and harden keypair edges

ananas-block:
## Changes

1. ECDH derivation-point guard (`is_derivation_point`) compares x-coordinates instead of full SEC1 encodings, so the negation of a committed point (`-P`: same x, flipped `0x03`/`0x02` parity byte) no longer bypasses it while yielding the same shared secret. This was the review's one open higher-severity item: `SigningKey::ecdh(-P_derive)` returned the P-256 derivation seed and `ViewingKey::ecdh(-P_pda)` the PDA nullifier root.
2. `PublicKey::confidential_view_tag` rejects the all-zero dummy owner explicitly, instead of relying on `as_p256`'s SEC1 re-parse to fail.
3. `ShieldedKeypair`'s `Signer::pubkey` / `sign_message` panic on a non-ed25519 keypair. Behavior change: a P-256/PDA keypair used through the infallible `Signer` API previously returned the system-program address (`Pubkey::default()`) and a zero signature.
4. `ShieldedPda::from_key_exchange` / `from_viewing_key` docs now state the two roots are disjoint and the constructors are not interchangeable.

## Tests

Negated committed points are rejected on both the `SigningKey` and `ViewingKey` ecdh rails; the zeroed owner is rejected by `confidential_view_tag` / `owner_proof_input_hash`; the P-256 `Signer` panic is pinned with `#[should_panic]`; and a non-degenerate encryption golden (non-palindromic salt, slot 1, block-crossing plaintext) pins the salt/slot wire encoding that the existing all-zero-salt / slot-0 golden did not.

## Verification

- `cargo test -p zolana-keypair` — 26 pass (13 integration + 9 HSM + 4 seed-based)
- `cargo clippy -p zolana-keypair --all-targets` and `cargo fmt --check` — clean
- `cargo check` on `zolana-transaction`, `zolana-wallet`, `zolana-client`, `zolana-program-test`, `zolana-cli` — all compile


ananas-block:
Opened from the wrong (already-merged) branch by mistake. Superseded by #249, which carries the actual changes. Closing.

PR #258, opened 2026-08-22
Custom Rings: RingRPC, SDK Integration, Custom Rings Generator

Atamanov:
This PR implements the audited custom ring, end to end. The SPP program is untouched. This is everything around the already merged ring rail.

Key decisions made:

- the ring program is upgradeable (and so is the auditor's key) - the audit ciphertext is per transaction, so it will only affect future txs;
- audit proving is integrated into the existing prover as a new custom-ring-audit circuit. It always runs through the Redis worker queue, sync requests are refused without it, and the health response lists the served circuits so probes can check for it;
- the auditor secret is kept only in ring-rpc, hosted mode derives per-ring keys from a root secret and signs them, `init` pins the service key;
- readers are on-chain records, reads are signed requests. A Solana key or a WebAuthn passkey is supported on the RPC as a reader. To be probably switched to compressed state instead of the PDA;
- a separate custom-ring-audit proof type keeps the SPP transfer circuits and their keys untouched, private_tx_hash as a shared public input binds the two proofs, and non-ring provers never load the audit key;
- zolana-ring cli is a helper CLI that allows you to easily create the custom ring template and select the features. A generated ring (template in the https://github.com/helius-labs/zolana-ring/), contains a copy of recent custom-ring/[program,sdk,tests].

Where is the logic:
- `custom-rings/program/src/instructions/transact.rs`, `verifier.rs`
- `prover/server/circuits/custom_ring`, `verifiable-encryption`
- `custom-rings/sdk/src/transfer.rs`, `PendingAuditProof` forces encrypt, then SPP proof, then audit proof, the hash order cannot be broken by accident
- `services/ring-rpc/src/audit.rs`, `keys.rs`, `webauthn.rs`
- `custom-rings/cli`.

To test:
`zolana-ring new`, then `zolana-ring pipeline` puts a live audited ring on devnet and reads the transfer back through the ring RPC. The TS SDK carries the ring scheme end to end.
https://d18al5hkzw7b42.cloudfront.net/ to interactively test.


sdk-libs/client/src/prover/inputs.rs
```diff
@@ -168,3 +169,75 @@ pub struct TransferP256Inputs {
...
+}
+
+const BN254_SCALAR_MODULUS: [u8; 32] = [
+    0x30, 0x64, 0x4e, 0x72, 0xe1, 0x31, 0xa0, 0x29, 0xb8, 0x50, 0x45, 0xb6, 0x81, 0x81, 0x58, 0x5d,
+    0x28, 0x33, 0xe8, 0x48, 0x79, 0xb9, 0x70, 0x91, 0x43, 0xe1, 0xf5, 0x93, 0xf0, 0x00, 0x00, 0x01,
+];
+
+fn canonical_audit_hash(value: [u8; 32]) -> Result<[u8; 32], ClientError> {
+    if value >= BN254_SCALAR_MODULUS {
+        return Err(ClientError::NonCanonicalField);
```
ananas-block: imo something like this is more descriptive, we cannot expect that devs are familiar with elliptic curves.
Ideally we don't expose the word field to devs at all.
```suggestion
        return Err(ClientError::GreaterThanBn254FieldSize);
```

prover/server/prover/common/circuit_utils.go
```diff
@@ -24,6 +24,14 @@ type BatchProofSystem struct {
 	ConstraintSystem constraint.ConstraintSystem
 }
 
+type Groth16ProofSystem struct {
```
ananas-block:
```suggestion
type RingProofSystem struct {
```

custom-rings/sdk/src/instructions/grant_reader/instruction.rs
```diff
@@ -0,0 +1,46 @@
+use solana_address::Address;
+use solana_instruction::{AccountMeta, Instruction};
+use zolana_interface::custom_ring::{tag, ReaderIxData};
+
+use crate::{shared::ReaderKey, CustomRing};
+
+#[must_use]
+pub struct GrantReader {
```
ananas-block: Why is it named `Grant` ?
I guess this is the instruction to configure read access right?

ananas-block: what about adding one instruction `configure_auditor_rpc_read_access` that uses an enum, this way we only need one additional instruction and it is easily extensible. 
see `pub enum UpdateProtocolConfigData {` as example.

custom-rings/program/src/instructions/create_config.rs
```diff
@@ -31,19 +24,33 @@ pub fn process_create_config_ix(accounts: &mut [AccountView], data: &[u8]) -> Pr
     let authority = iter.next_signer("authority")?;
     let config_account = iter.next_mut("config")?;
     let system_program = iter.next_account("system_program")?;
+    let program = iter.next_account("program")?;
+    let program_data = iter.next_account("program_data")?;
```
ananas-block, line 32: it should be guarded by the configured authority not the upgrade authority.
The program upgrade authority should only be used to initialize the config after initial deployment.

custom-rings/sdk/src/instructions/grant_reader/instruction.rs
ananas-block: hm ok since we create new accounts this makes less sense in this case.

program-libs/interface/src/custom_ring.rs
ananas-block, line 1: this file doesn't belong here. The consistent pattern would be to have an interface crate in `custom-rings`.
Please either move it into it's own crate in `custom-rings` or into the `custom-rings/program`.

custom-rings/interface/src/verifying_key.rs
ananas-block, line 1: this file doesn't belong here. The consistent pattern would be to have an interface crate in custom-rings.
Please either move it into it's own crate in custom-rings or into the custom-rings/program.

ananas-block: also let's rename `auditor_key_encryption` to something better like `custom_ring_policy_proof` or similar.
(I realize that naming originates from my pr xD)

sdk-libs/client/src/solana_rpc.rs
```diff
@@ -62,6 +62,15 @@ pub struct ConfirmedInstructionGroups {
 
 const DEFAULT_CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(30);
 
+impl ConfirmedInstructionGroups {
+    pub fn from_confirmed_transaction(
```
ananas-block: can we implement `TryFrom` instead?

sdk-libs/client/src/prover/client.rs
```diff
@@ -240,6 +243,14 @@ impl ProverClient {
         self.send(to_json_p256_ring(inputs), self.delivery)
     }
 
+    pub fn prove_auditor_key_encryption(
```
ananas-block: I think it would be cleaner if we had a clear dependency hierarchy so that all custom ring crates depend on SPP crates but SPP crates must depend on or include any logic that is only used for custom rings.
Maybe we will need an exception in the keypair crate.
Maybe it makes sense to say every crate in sdk-libs only contains SPP logic and contain any logic for custom rings to `custom-rings`.

sdk-libs/transaction/src/instructions/transact/transfer.rs
```diff
@@ -409,6 +419,38 @@ fn validate_settlement_target(
 }
 
 impl PreparedTransfer {
+    pub fn compact_outputs(mut self) -> Result<Self, TransactionError> {
```
ananas-block: this feels hacky, why do we need this?

sdk-libs/ts/src/wallet/transaction-slots.ts
ananas-block, line 1: why `parties.ts` ?

sdk-libs/client/src/prover/inputs.rs
Atamanov: Moved the audit hash types out of the client into custom-ring-sdk, the check now returns AuditProofInputError::HashOutOfRange (error renamed to omit the ec terminology)

prover/server/prover/common/circuit_utils.go
Atamanov: Renamed to RingProofSystem, with GetRingSystem, writeRingProofSystem

custom-rings/sdk/src/instructions/grant_reader/instruction.rs
Atamanov: Renamed the surface to say what it configures. Now reads as GrantReadAccess, RevokeReadAccess. The account is ReadAccessRecord, TS mirrors changed accordingly. On the single enum instruction: agree with the followup.

custom-rings/program/src/instructions/create_config.rs
Atamanov: The config authority is now mutable: SET_AUTHORITY (tag 6), guarded by the current configured authority.

program-libs/interface/src/custom_ring.rs
Atamanov: Moved to a new custom-rings/interface crate (custom-ring-interface), split into state / instruction / audit / audit_vk modules. zolana-interface is back to SPP-only.

custom-rings/interface/src/verifying_key.rs
Atamanov: Done both. This file now lives in the new custom-rings/interface crate (custom-ring-interface). Split into state, instruction, audit and audit_vk modules, with its own vk fingerprint pin.

sdk-libs/client/src/prover/client.rs
Atamanov: Done

sdk-libs/ts/src/wallet/transaction-slots.ts
Atamanov: 🕺🕺🕺 xD  parties.ts -> transaction-slots.ts

sdk-libs/transaction/src/instructions/transact/transfer.rs
Atamanov: Replaced. The layout is chosen at prepare time now.

sdk-libs/client/src/solana_rpc.rs
Atamanov: Done

sdk-libs/keypair/Cargo.toml
```diff
@@ -8,11 +8,13 @@ repository.workspace = true
 
 [dependencies]
 p256 = { workspace = true }
+bs58 = { workspace = true }
 ed25519-dalek = { workspace = true }
 aes = { workspace = true }
 ctr = { workspace = true }
 hkdf = { workspace = true }
 zeroize = { workspace = true }
+zolana-interface = { path = "../../program-libs/interface", default-features = false }
```
ananas-block: I would rather have duplicate constants than this dependency tbh
would it make sense to move the constants into the hasher crate?
Or we just have duplicate constants in this crate and the ring interface.

Atamanov: moved the constants into the hasher crate

ananas-block:
I think we should make the aggregate proof/circuit naming generic before merging.

Every custom ring will always have an auditor, so auditing is a foundational invariant—not an optional feature. Names such as `auditor_pubkey`, `AuditorMessage`, `AuditEncryption`, audit-specific errors, reader access, and Ring RPC audit APIs are therefore correct.

However, the current proof is the complete custom-ring policy proof. We expect to extend that same proof/circuit with additional configurable constraints, so names identifying the aggregate proof as solely “audit” are too narrow.

The naming rule should be: anything representing the aggregate proof, circuit, public input, or proving pipeline uses `CustomRing*`; audit names remain only for the underlying auditor-specific machinery.

Suggested updates:

1. `AuditProof` → `CustomRingProof`.
2. `CustomRingTransactIxData.proof: AuditProof` → `proof: CustomRingProof`.
3. `AuditPublicInput` → `CustomRingPublicInput`.
4. `AuditPublicInputHash` → `CustomRingPublicInputHash`.
5. `AuditPrivateTxHash` → `CustomRingPrivateTxHash`.
6. `AuditProofParams` → `CustomRingProofParams`.
7. `PendingAuditProof` → `PendingCustomRingProof`.
8. `AuditProofRequest` → `CustomRingProofRequest`.
9. `AuditProofRequestJson` → `CustomRingProofRequestJson`.
10. `AuditProofError` → `CustomRingProofError`.
11. `AuditProofInputError` → `CustomRingProofInputError`.
12. `canonical_audit_hash` → generic custom-ring/field-hash validation naming.
13. `RingTransactWithAudit` → `CustomRingTransact`.
14. `RingTransactWithAudit.audit_proof` → `proof`.
15. `AuditedTransfer` → `CustomRingTransfer`.
16. `AuditedTransferInput` → `CustomRingTransferInput`.
17. `ProvenTransfer.audit_proof` → `proof` or `ring_proof`.
18. Aggregate interface module `audit.rs` → `proof.rs`, `circuit.rs`, or another generic name.
19. `audit_vk` → `custom_ring_vk` or `verifying_key`.
20. References such as `audit_vk::VERIFYINGKEY` → corresponding generic path.
21. Circuit directory `circuits/custom_ring/audit` → `circuits/custom_ring` or `circuits/custom_ring/transfer`.
22. Go package `audit` → generic custom-ring circuit package.
23. The aggregate Go `Circuit` should live under that generic package; optionally rename it `CustomRingCircuit`.
24. `CustomRingAuditCircuitType` → `CustomRingCircuitType`.
25. Wire circuit identifier `"custom-ring-audit"` → `"custom-ring"` or `"custom-ring-transfer"`.
26. `CustomRingAuditParameters` → `CustomRingParameters`.
27. `customRingAuditParametersJSON` → `customRingParametersJSON`.
28. `R1CSCustomRingAudit` → `R1CSCustomRing`.
29. `SetupCustomRingAudit` → `SetupCustomRing`.
30. `ConvertCustomRingAudit` → `ConvertCustomRing`.
31. `ProveCustomRingAudit` → `ProveCustomRing`.
32. `setup-custom-ring-audit` CLI command → `setup-custom-ring`.
33. `convert-custom-ring-audit` → `convert-custom-ring`.
34. Enabled-circuit value `custom-ring-audit` → generic circuit identifier.
35. `custom_ring_audit_transfer.key` → `custom_ring_transfer.key` or a generic versioned name.
36. Update the proving-key lockfile and generation scripts to use the generic artifact name.
37. `CustomRingAuditQueueWorker` → `CustomRingQueueWorker`.
38. `NewCustomRingAuditQueueWorker` → `NewCustomRingQueueWorker`.
39. `getCustomRingAuditMaxConcurrency` → `getCustomRingMaxConcurrency`.
40. `processCustomRingAuditProof` → `processCustomRingProof`.
41. Server handler `customRingAuditProof` → `customRingProof`.
42. `errCustomRingAuditProof` → `errCustomRingProof`.
43. `zk_custom_ring_audit_queue` → `zk_custom_ring_queue`.
44. `zk_custom_ring_audit_processing_queue` → `zk_custom_ring_processing_queue`.
45. `CUSTOM_RING_AUDIT_WORKER_CONCURRENCY` → `CUSTOM_RING_WORKER_CONCURRENCY`.
46. Metrics, health responses, routing labels, and error prefixes using `custom-ring-audit` → generic circuit naming.
47. `sdk/tests/audit_circuit.rs` → `custom_ring_circuit.rs` when it tests the aggregate circuit.
48. `sdk/tests/audit_proof_params.rs` → `custom_ring_proof_params.rs`.
49. Whole-circuit Go tests currently under the `audit` package → generic custom-ring circuit tests.
50. Comments describing the complete proof as the “audit proof” or complete circuit as the “auditor circuit” → “custom-ring proof/circuit with mandatory audit constraints.”
51. README wording that says this particular ring “adds an auditor” should instead state that auditing is foundational to every custom ring.

Names such as `auditor_pubkey`, `AuditorMessage`, `AuditEncryption`, `EncryptedAudit`, `AUDIT_CIPHERTEXT_LEN`, `AUDITOR_MESSAGE_LEN`, `RingAudit`, reader authorization, and auditor-specific errors should remain unchanged.

I’m not suggesting that we make the account layouts or instruction formats extensible in this PR. Breaking changes there can happen when additional configuration is introduced. This is only about avoiding an unnecessarily narrow identity for the proof and circuit now.

ananas-block:
One additional naming cluster exists outside the proof/circuit layer: the complete custom-ring transfer flow is also named after auditing in parts of the TypeScript SDK and supporting documentation.

Suggested updates:

1. `AuditedTransferParams` → `CustomRingTransferParams`.
2. `proveAuditedTransfer` → `proveCustomRingTransfer`.
3. Update `buildRingTransferTransaction` to call `proveCustomRingTransfer`.
4. Update `buildRingWithdrawalTransaction` to call `proveCustomRingTransfer`.
5. Update the corresponding exports from `sdk-libs/ts/src/ring/index.ts`.
6. Change comments saying the TypeScript implementation mirrors Rust `AuditedTransfer` to `CustomRingTransfer`.
7. `EncryptedAuditedTransfer` → `EncryptedCustomRingTransfer`.
8. `WalletAuthority.encryptAuditedTransfer` → `encryptCustomRingTransfer`.
9. Rename the corresponding wallet-authority implementation method.
10. Update exports of `EncryptedAuditedTransfer` from the transaction and wallet index modules.
11. Update TypeScript tests and fixtures calling `encryptAuditedTransfer` or `proveAuditedTransfer`.
12. Replace “audited transact” with “custom-ring transact” when referring to the complete transaction.
13. Replace “audited instruction” with “custom-ring instruction.”
14. Replace “audited transfer” with “custom-ring transfer” in CLI help and the custom-ring README.
15. Rename CI label `custom-ring auditor lifecycle` → `custom-ring lifecycle`.

`AuditWitness`, its `audit` field, `AuditorMessage`, and auditor-view result types should remain audit-specific because they describe the mandatory auditor mechanism or its decrypted output.

ananas-block:
A final deduplicated set of aggregate naming occurrences not explicitly covered by the previous two comments:

1. `AUDIT_PROOF_LENGTH` → `CUSTOM_RING_PROOF_LENGTH`.
2. `CompressedProof.toAuditProof()` → `toCustomRingProof()`.
3. TypeScript `customRingAuditRequest()` → `customRingProofRequest()`.
4. Both TypeScript client methods named `proveCustomRingAudit()` → `proveCustomRingProof()`.
5. `checkedAuditProof()` → `checkedCustomRingProof()`.
6. `RING_AUDIT_PROOF_LENGTH` → `RING_PROOF_LENGTH`.
7. Rust `TransferError::AuditInput` and `TransferError::AuditProof` → `ProofInput` and `Proof`.
8. Remove “minimal” and “example” terminology from the canonical custom-ring program documentation and Justfile descriptions.

Audit-specific encryption, witness, message, reader, and decrypted-view names remain correctly scoped and do not need changing.

ananas-block:
One additional naming cluster remains that is not covered by the previous comments: the TypeScript SDK still uses legacy **zone** terminology for what Rust and the new APIs call a **ring**. Since this PR is actively adding to these APIs (notably `withZoneProgramId()` whose documentation already says it binds the note to a ring), I think this should be made consistent now:

1. `UtxoInit.zoneProgramId` -> `ringProgramId`.
2. `Utxo.zoneProgramId` -> `ringProgramId`.
3. `ProofInputUtxo.zoneDataHash` -> `ringDataHash`.
4. `ProofOutputUtxo.zoneProgramId` / `zoneDataHash` -> `ringProgramId` / `ringDataHash`.
5. `ProofOutputInit.zoneProgramId` / `zoneDataHash` -> the ring equivalents.
6. `withZoneProgramId()` -> `withRingProgramId()`.
7. `resolveZoneProgramId()` -> `resolveRingProgramId()`.
8. `DataRecord` kind `"zoneData"` -> `"ringData"`.
9. `Data.zoneData()` -> `Data.ringData()`.
10. Transaction error names containing `*_ZONE_*` -> `*_RING_*`.
11. Persisted wallet fields `zoneProgramId` / `zoneDataHash` -> the ring equivalents.
12. Serialization-codec fields, parameters, and local variables using `zone*` -> `ring*`; this does not require changing the binary layout.
13. Wallet sync, merge, actions, and private-transaction internals should use `ring*`.
14. Ring transfer code should call `withRingProgramId()` and read `utxo.ringProgramId` instead of translating back into zone terminology.
15. Audit decoding should no longer map `plaintext.zoneProgramId` into `ringProgramId`; both sides should use `ringProgramId`.
16. Update the corresponding TypeScript tests, fixtures, and documentation.

After this cluster, I do not see another high-confidence systemic naming problem. The remaining `Audit*` names mostly describe genuinely auditor-specific services, ciphertexts, witnesses, messages, or decrypted views.

ananas-block:
`createAuditorKey` should require an authenticated request signed by either the program’s current Loader v3 upgrade authority or the authority stored in the custom-ring config. The signature should bind the cluster/genesis hash, `ringProgramId`, timestamp, and nonce, and Ring RPC should verify the relevant on-chain authority before returning the attested auditor key.

Atamanov:
Renames ✅ , auth on RingRPC ✅. Please check.

ananas-block:
Hosted-RPC rerun remains broken. The first run writes `auditor_pubkey_file`, so the second run hits `LocalAuditorAgainstRemoteRpc` before the existing config is read and the new `Chain` arm is reachable. Please prioritize `Some(config) => Chain(config.auditor_pubkey)` and apply the guard only to `(None, true)`.

ananas-block:
`emulatedFpToBytes` uses `field.Reduce`, which permits the 256-bit representation `x+p`. A modified prover can grind an ECDH output where `x+p` fits, derive the ciphertext from those noncanonical bytes, and produce a valid proof that the auditor cannot decrypt using canonical `x`.

ananas-block:
`init` loads `config_authority_keypair` via `funded_authority()` and uses it to sign `CREATE_CONFIG`, but the program requires the Loader v3 upgrade authority. Initialization therefore fails whenever `config_authority_keypair` and `upgrade_authority_keypair` are different.

ananas-block:
`prove` overwrites every output’s `ring_program_id` before validation, making output-side `ForeignRing` unreachable and silently rebinding default outputs to the custom ring, contrary to the README. Please preserve output membership and add functional custom→default and default→custom transfer tests; current lifecycle coverage only exercises SOL deposit and custom→custom transfers. Feel free to fix this in a separate PR, but please document it here as a TODO.

ananas-block (dismissed):
feel free to merge to unblock a handoff release, I am ok with fixing the last comments in a follow on pr

ananas-block (dismissed):

alex-helius (approved):

alex-helius (approved):

PR #264, opened 2026-08-25
Add plaintext compressed-account example

ananas-block:
## Summary

  - Adds a plaintext compressed-account example that uses a deterministic nullifier as the account’s stable address.
  - Implements separate create and update instructions. Creation proves the address does not yet exist; updates spend the current state while preserving the address.
  - Uses an incrementing version counter as the UTXO blinding value, producing a new commitment for each update.
  - Publishes account state without encryption and provides SDK support for proof construction, state discovery, decoding, and validation.
  - Adds end-to-end coverage for creation, updates, duplicate creation, and stale-update rejection.


tilo-14 (approved):
maybe add a readme

ananas-block:
will add a readme with a description of what to look out for when using SPP for compression in a separate pr.

PR #267, opened 2026-08-25
feat: remote wallet authorities, verified prover responses, and one wallet scan

sergeytimoshin:
Supersedes #228, which is closed. Its four commits are the first commit here, unchanged in content; everything after builds on them. `keypair/turnkey-backend` is rebased onto this branch and stacks cleanly.

**Motivation:**
1. A backend whose signing key lives in an HSM, a KMS, or a remote custodian could not become a wallet authority. `ShieldedKeypairTrait` had no consumer, both authority impls were hardwired to `ShieldedKeypair`, and the four encryption bodies a hand-rolled `SyncWalletAuthority` needs were private free functions — so a backend either duplicated protocol-critical glue or held a throwaway local `ShieldedKeypair`, detaching the published identity from the device key.
2. `RoleExpansion` was `pub(crate)`, so an out-of-crate backend could not expand a device-produced seed without reimplementing the HKDF schedule. A divergent expansion yields a valid-looking *different* identity: signatures verify, addresses look right, nothing catches it.
3. A prover response was trusted on arrival. A remote prover could return a proof that fails on-chain, and the wallet would learn that only from a rejected transaction.
4. A tag scan could not advance past an empty indexer page, so a wallet rescanned the same empty range forever.
5. `sync_parallel` was a second copy of the scan body, and the copy had drifted.

**Changes:**

*Authority surface (was #228)*
1. `derivation::expand_roles(seed, curve)`, one public entry for both role secrets. Validates seed width per rail (64 ed25519, 32 P-256): HKDF-Extract accepts any length, so an unchecked truncated seed expands into that same wrong-but-valid identity. `KeypairError` becomes `#[non_exhaustive]` since out-of-crate backends match on it. `RoleExpansion` stays private — publishing it means publishing `Rail` alongside `Curve`, two overlapping public enums for one concept.
2. `KeypairWalletAuthority` generic over the keypair, `K = ShieldedKeypair` the way `HashMap` defaults its hasher. `with_viewing_keys` takes current plus historical keys, which the old constructor could not express, and is fallible with its own `AuthorityViewingKeyMismatch`: encryption keys off the keypair while a scan keys off the vector, so a mismatched pair would encrypt to one key and scan with another.
3. `forward_viewing_key_trait!` at all four forwarding sites. Not a blanket impl over an accessor trait: coherence would stop a downstream crate implementing `ViewingKeyTrait` by hand, which is exactly what a device-resident viewing key needs.
4. `LocalWalletAuthority` -> `KeypairWalletAuthority`, Rust and TypeScript. Generalizing the type made the old name false. No alias.

*New in this branch*
5. `ClientEd25519WalletAuthority`: privacy roles derived from a remote signer's deterministic derivation signature. Holds no Solana signing key — it verifies the signature, expands the roles locally, and serves sync, transaction construction, and the prover. The final transaction is authorized separately by the signer, which is why its approval hook is an explicit documented no-op rather than an inherited default.
6. Prover responses are verified locally before they can influence a signing request, on all five client proof paths, against the committed shape-specific verifying key and the locally computed public input. Redirects are disabled on the prover HTTP clients.
7. `scannedThrough` on both tag endpoints, so a terminal page advances the scan even when nothing matched. Photon reports the stream frontier inside the same `REPEATABLE READ` transaction as the page, and only when the limit did not truncate.
8. Self transfers are classified consistently. A transfer whose every real recipient is this wallet is a self transfer on every rail — the anonymous rail hardcoded `Outbound` while its own recipient receipt already said `SelfTransfer`, so one row contradicted the other. Self-recognition covers every viewing key the wallet has held, not just the current one; a self transfer addressed to a rotated-out key was classified `Outbound`, reachable only through the new historical-key support. A rescan replaces a refined row instead of adding a second beside it.
9. One scan, generic over a `ProbeFanout` strategy. `sync_parallel`'s copy had lost `record_confidential_send` entirely — it stored a confidential send's UTXOs but recorded no outbound history for it — and the cursor pruning for spent nullifiers. Both entry points survive so the benchmark can still compare them, but they can no longer disagree. `parallel.rs` drops from 267 lines to 60 with no scan logic left in it.

**Testing:**
1. `a_remote_keypair_backend_is_a_wallet_authority` drives a double that is deliberately not a `ShieldedKeypair`; identity, signature and `tx_viewing_pk` pinned to the software authority, since equality against the software path is the only assertion that catches a wrong-but-well-formed identity. `expand_roles` pinned against `from_keypair` on both rails, comparing secrets rather than pubkeys, plus wrong-width and PDA cases.
2. Both classification rules pinned in Rust and TypeScript, and each was confirmed to fail before its fix. Serial and parallel are compared on the confidential and anonymous rails.
3. `build_unified_transfer` placed the recipient at output slot 1, inside the `SENDER_SLOT_COUNT` range the protocol reads as sender change, which netted the sender's history row to zero — which is why no history test could be written against it. Recipients now start after the sender slots.
4. `just test-sdk-libs`, `just clippy`, `just check-all`, `cargo fmt --all --check`. The client proving matrix passes 9/9.
5. `npm run check:ts`: oxfmt, oxlint, `tsc --noEmit`, 273 vitest tests, build.

**Notes:**
1. `parallel` is off by default, so `--workspace` compiles neither `wallet::parallel` nor the tests holding the two strategies to identical results. Added to `test-sdk-libs`, `clippy` and `check-all` — without it, the drift this PR fixes can recur unnoticed.
2. TS fixtures are sha256-verified against their manifest and `cargo` passes regardless, so `npm run check:ts` is needed for anything touching `sdk-libs/ts`, including Rust-looking changes.
3. This is a large PR because it absorbs #228. The natural split is at `feat: add client-owned Ed25519 wallet authority`: the authority surface below, the sync/prover/indexer work above. Happy to split if you would rather review it that way.


ananas-block:
This duplicates the shape → VK mapping in `CircuitId::verifying_key()`. Please use `CircuitId::ConfidentialEddsa(u8::try_from(n_in)?, u8::try_from(n_out)?, CURRENT_PUBLIC_ASSET_SLOTS).verifying_key()` to avoid another synchronization point.

ananas-block:
Rust uses a stringly `Authority` error here, while TypeScript reports the unrelated `KEYPAIR_INVALID_SECRET_KEY`. Please add a named `InvalidDerivationSeed` error on both sides so the failure is precise and consistent.

ananas-block:
This only tests the guard in the `RemoteKeypair` test double; `sign_p256_with` merely forwards the payload. It would pass with no library guard, so please test library-owned behavior or remove/reframe the claim that the guard survives authority indirection.

sergeytimoshin:
All three addressed in f7b26c5e.

1. **VK lookup** — now `CircuitId::ConfidentialEddsa(n_in, n_out, N_PUBLIC_SLOTS as u8).verifying_key()`, so the client resolves through the same dispatch the on-chain verifier uses and the local table is gone. `verifying-keys` is already a default feature of `zolana-interface`, so no Cargo change was needed. Worth noting how I missed it: I did grep for `transfer_confidential_1_1` while removing a *test*-local copy of the same table, but filtered with `grep -v "verifying_keys/"` on the assumption that directory held only generated constants. That hid `circuit.rs`, so I deleted one duplicate while leaving a second and called it canonical.

2. **Seed errors** — added `TransactionError::InvalidDerivationSeed` and `TRANSACTION_INVALID_DERIVATION_SEED`, replacing the stringly `Authority` error and the unrelated `KEYPAIR_INVALID_SECRET_KEY`. Kept distinct from `KeypairError::InvalidDerivationSeed`, which is the wrong-width case; both are documented against each other. I had spotted this mismatch earlier and talked myself out of it as pre-existing convention, which was wrong: the width check I added in the same PR is what made the verification path the odd one out.

3. **Guard test** — you're right that it proved only the double guards itself. Reframed as `derivation_shaped_payloads_are_refused_before_signing`, wrapping a real `ShieldedKeypair` so `SigningKey::sign_hash` is what's under test, with the reasoning recorded in a doc comment so it doesn't drift back. Also asserts a non-derivation payload still signs, so it can't pass by signing never working. Verified load-bearing: deleting the guard in `signing_key.rs` fails the test.

Verified: `just clippy` (including the `--features parallel` pass), `cargo fmt --check`, tests green on `zolana-{transaction,keypair,wallet,client}` and under `--features parallel`; client proving 5/5 including the all-shapes sweep, which now exercises `CircuitId` for every shape; TS 273 tests, lint and typecheck clean.

ananas-block (approved):

PR #268, opened 2026-08-25
docs: update architecture diagram

ananas-block:
## Summary

- Rename the "Policy Program" cells to "Custom Ring Program" and reword the SPP rowspan comment to match the current ring vocabulary.
- Add a "Transaction RPC" row (optional relayer for anonymous tx) to the Client and RPC tables. The client now sends transactions through it; the RPC submits confidential transactions directly to SPP and relays anonymous transactions so the user is not the fee payer. The previous direct Client -> SPP / ZK Swap / Policy Program N edges are removed.
- Regenerate `architecture.png` and `architecture.svg` from the updated DOT source.
- Add a `just render-diagrams` recipe that renders every `docs/diagrams/*.dot` to PNG (144 dpi) and SVG.

## Test plan

- [ ] `just render-diagrams` reproduces the committed PNG/SVG from the DOT source.
- [ ] Rendered diagram shows the Transaction RPC row with both the relayed anonymous and confidential edges into SPP.


tilo-14 (approved):
lgtm

PR #272, opened 2026-08-27
ci: fix the flaky tests and prebuild the localnet matrix

Atamanov:
Fixes the three flaky CI tests and removes repeated build work from the pipeline.

- `ring-rpc` unit skew tests take one clock reading through `decide_at`, boundary timestamps stay outside the window under any load.
- The hub takes a `Clock` (`System` in production, `Fixed` in tests), the audit service test pins time and asserts the exact window, ±60 s accepted, ±61 s refused, replay eviction at the edge.
- `dynamic-swap` escrow tests poll photon with a 30 s deadline (`wait_until`) after each send, note selection waits for indexing.
- A `build / localnet suites` job compiles the programs, the cli, xtask, the prover, and nine nextest archives once, the localnet matrix jobs run from the archives with zero compilation (`tools/ci/nextest-suite.sh`, plain `cargo nextest run` when `ZOLANA_NEXTEST_ARCHIVE_DIR` is unset).
- The photon binary is cached by the hash of its source closure, pushes outside it skip the build.
- The lean cache gets a `restore-keys` prefix, a manifest bump rebuilds incrementally.

### DIFF
--- a/.github/workflows/formal-verification.yml
+++ b/.github/workflows/formal-verification.yml
@@ -52,6 +52,8 @@ jobs:
         with:
           path: prover/server/formal-verification/.lake
           key: lean-${{ hashFiles('prover/server/formal-verification/lean-toolchain') }}-${{ hashFiles('prover/server/formal-verification/lake-manifest.json') }}
+          restore-keys: |
+            lean-${{ hashFiles('prover/server/formal-verification/lean-toolchain') }}-
 
       - name: Build Lean project (checks all theorems)
         run: |
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -204,10 +204,47 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+  build-localnet-tests:
+    name: build / localnet suites
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    env:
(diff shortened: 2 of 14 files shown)


sergeytimoshin:
correctness nitpicks:

   1. The rfq leg still compiles from source — "zero compilation" doesn't hold. test-rfq-validator runs plain cargo test -p rfq-test --test rfq (justfile:1360), not
   nextest-suite.sh, and rfq-test has no archive in build-localnet-archives. In the matrix job with ZOLANA_PREBUILT=1 that leg will build rfq-test and whatever deps the shared
   rust cache didn't carry. 

   2. custom-ring leg also compiles. cargo build -q -p custom-ring-cli at justfile:1238 runs unconditionally — it's not guarded by ZOLANA_PREBUILT and the cli isn't in the uploaded artifacts. 

   3. Photon cache key misses two inputs. I traced the path-dep closure of photon-indexer: zolana-event, zolana-indexer-api, zolana-interface, zolana-tree → batched-merkle-tree,
   hasher → account-checks, bloom-filter — all covered by the hash list, plus Cargo.lock and the toolchain, so the "source closure" claim is accurate for crate sources. But the
   root Cargo.toml (release profile, workspace dependency features) and .cargo/config.toml (rustflags) both change the binary without invalidating the cache. Add both to the
   hashFiles list.

sergeytimoshin (approved):

PR #273, opened 2026-08-27
custom-ring: prove a transfer over an async transport

sergeytimoshin:
`CustomRingTransfer::prove` is blocking-only, and the only Solana `Rpc` implementation lives behind `zolana-client`'s `solana-rpc` feature. A host that cannot link that feature cannot prove a ring transfer at all, even though it already speaks `AsyncRpc` over its own transport.

The rest of this SDK is async-first -- `sync_wallet_async`, `create_transfer`, `AsyncProverClient` -- so the ring transfer was the outlier.

## Change

Adds `prove_async` over `AsyncRpc` and `AsyncProverClient`, **alongside** the blocking path rather than replacing it. Every existing call site is untouched; the CLI and the ring test validator build unchanged.

Five reads separate the two paths:

1. the ring config,
2. the tree's dummy-input policy,
3. the indexer's inclusion proofs,
4. the indexer's non-inclusion proofs,
5. the two prover calls.

Everything between them is proof assembly. Duplicating it would have let the two paths drift on the ordering the auditor message depends on -- the comment in `prove` is emphatic about it for good reason. So the reads move to the edges and the middle is shared:

- `AccountRead::decode` -- validating a fetched account is transport-independent
- `dummy_input_policy` -- likewise for the tree account
- `RingSpendInputs::{queries, assemble}` -- what to ask for, and how to pair the answers
- `StagedTransfer` -- the work between the reads

`StagedTransfer` takes the pending proof out rather than borrowing it, so `finish` cannot run before `proof_request` has bound the real `private_tx_hash`. That is the same invariant `PendingCustomRingProof` already encodes, kept across the split.

## Verified

- builds and tests with default features and with `--no-default-features`
- `custom-ring-cli` and `custom-ring-test-validator` build untouched
- fmt and clippy clean

Also checked against the actual consumer: with this branch patched in, a QOS enclave wallet -- which cannot link `solana-rpc`, because that pulls tokio past the libc version `qos_core` pins -- compiles a call to `prove_async` using its own `AsyncRpc` transports. That was the motivating case, and it does not build without this.

## Why

Building custom-ring support into a TVC enclave wallet. #270 made the crate linkable there; this makes the transfer path usable there.

### DIFF
--- a/custom-rings/sdk/src/transfer.rs
+++ b/custom-rings/sdk/src/transfer.rs
@@ -43,7 +43,7 @@ const NO_RING_DATA_HASH: [u8; 32] = [0u8; 32];
 #[must_use = "prove or discard the transfer explicitly"]
 pub struct CustomRingTransfer<'a> {
     ring: CustomRing,
-    sender: &'a dyn ViewingKeyTrait,
+    sender: &'a (dyn ViewingKeyTrait + Send + Sync),
     prepared: PreparedTransfer,
     interface_transfer_accounts: Vec<TransactInterfaceTransferAccounts>,
     tree: Option<Address>,
@@ -59,7 +59,11 @@ pub struct CustomRingTransferInput<'a> {
     /// owner signs the assembled Solana transaction.
     ///
     /// `ShieldedKeypair` implements the trait, so passing one still works.
-    pub sender: &'a dyn ViewingKeyTrait,
+    ///
+    /// `Send + Sync` so that [`Self::prove_async`]'s future is `Send`. Without
+    /// it the async path cannot be awaited on a multi-threaded runtime, which
+    /// is exactly where a host that needs the async path runs.
+    pub sender: &'a (dyn ViewingKeyTrait + Send + Sync),
     pub prepared: PreparedTransfer,
 }
 
@@ -883,6 +887,35 @@ mod tests {
             )],
         )
         .expect("withdrawal accounts");
+    }
+
+    /// Every `AsyncRpc` method has a default, so an empty type is a valid one.
+    struct NoRpc;
+    impl AsyncRpc for NoRpc {}
+
+    #[test]
+    fn the_async_prove_future_is_send() {
+        // A host reaches for the async path because it runs on a multi-threaded
+        // runtime; a future that cannot cross threads is no use to it. The
+        // `Send + Sync` bound on `sender` is what makes this hold, and dropping
+        // it fails here rather than in whatever server tries to await it.
+        fn assert_send<F: Send>(_: F) {}
+
+        let (keypair, prepared) = prepared_transfer(4);
+        let prover = AsyncProverClient::new(String::new());
+        let rpc = NoRpc;
+        assert_send(
+            CustomRingTransfer::new(CustomRingTransferInput {
+                ring: ring(),
+                sender: &keypair,
+                prepared,
+            })
+            .prove_async(AsyncTransferProofEnvironment {
+                indexer: &rpc,
+                rpc: &rpc,
+                prover: &prover,
+            }),
+        );
     }
 
     #[test]
--- a/custom-rings/sdk/src/transfer.rs
+++ b/custom-rings/sdk/src/transfer.rs
@@ -1,13 +1,14 @@
 use rand::{rngs::OsRng, RngCore};
+use solana_account::Account;
 use solana_address::Address;
 use solana_instruction::Instruction;
 use solana_signature::Signature;
 use solana_signer::Signer;
 use thiserror::Error;
 use zolana_client::{
-    ClientError, ProofCompressed, ProverClient, RingTransferProofResult, RingTransferProver, Rpc,
-    SettlementAccountValidation, Shape, SpendProof, SppProofInputUtxo, SppProofInputs,
-    TransferSpendInput,
+    AsyncProverClient, AsyncRpc, ClientError, MerkleProof, NonInclusionProof, ProofCompressed,
+    ProverClient, RingTransferProofResult, RingTransferProver, Rpc, SettlementAccountValidation,
+    Shape, SpendProof, SppProofInputUtxo, SppProofInputs, TransferSpendInput,
 };
 use zolana_interface::event::OutputDataEncoding;
 use zolana_interface::{
@@ -19,7 +20,8 @@ use zolana_interface::{
     N_PUBLIC_SLOTS, SHIELDED_POOL_PROGRAM_ID,
 };
 use zolana_keypair::{
-    random_blinding, random_salt, KeypairError, ShieldedKeypair, ViewingKey, ViewingKeyTrait,
+    random_blinding, random_salt, KeypairError, P256Pubkey, ShieldedKeypair, ViewingKey,
+    ViewingKeyTrait,
 };
 use zolana_transaction::{
     instructions::transact::{
@@ -32,7 +34,8 @@ use zolana_tree::{TreeAccount, TreeError};
 
 use crate::{
     to_instruction_proof, AccountReadError, CustomRing, CustomRingProof, CustomRingProofError,
-    CustomRingProofInputError, CustomRingProofParams, CustomRingTransact, Deposit, EncryptedAudit,
+    CustomRingProofInputError, CustomRingProofParams, CustomRingProofRequest, CustomRingTransact,
+    Deposit, EncryptedAudit, PendingCustomRingProof,
 };
 
 const NO_RING_DATA_HASH: [u8; 32] = [0u8; 32];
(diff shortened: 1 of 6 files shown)


custom-rings/sdk/src/transfer.rs
```diff
@@ -717,6 +889,35 @@ mod tests {
...
+                prepared,
+            })
+            .prove_async(AsyncTransferProofEnvironment {
+                indexer: &rpc,
+                rpc: &rpc,
+                prover: &prover,
+            }),
+        );
+    }
+
```
Atamanov, line 985: This test only constructs the future and verifies Send. It never polls it or exercises an async RPC/prover call. Please consider executing one complete prove_async transfer in the ring lifecycle suite and assert parity with the synchronous one.

custom-rings/sdk/src/transfer.rs
```diff
@@ -241,51 +322,90 @@ impl<'a> CustomRingTransfer<'a> {
...
+    /// recomputes that same public-input chain from the payload and the config
+    /// account.
+    fn proof_request(
+        &mut self,
+        ring_result: RingTransferProofResult,
+    ) -> Result<(CustomRingProofRequest, RingTransferProofResult), TransferError> {
+        let pending = self
+            .pending_proof
+            .take()
+            .ok_or(TransferError::MissingRingConfig)?;
```
Atamanov: looks like unrelated error to the case

custom-rings/sdk/src/transfer.rs
```diff
@@ -241,51 +322,90 @@ impl<'a> CustomRingTransfer<'a> {
...
+            ring: self.ring,
+        })
+    }
+}
+
+/// A transfer past validation and auditor encryption, waiting on the reads.
+struct StagedTransfer {
+    tx_viewing_key: ViewingKey,
+    /// Taken by [`Self::proof_request`]; `finish` cannot run before it does.
+    pending_proof: Option<PendingCustomRingProof>,
```
Atamanov: This encodes two protocol states in one StagedTransfer. finish is callable in either state because it never inspects pending_proof. A repeated proof_request reports the unrelated MissingRingConfig error. Consider consuming StagedTransfer in proof_request and returning a finish-only type with no pending_proof.

custom-rings/sdk/src/transfer.rs
```diff
@@ -183,17 +197,85 @@ impl<'a> CustomRingTransfer<'a> {
...
+            .read_config_async(environment.rpc)
+            .await?
+            .ok_or(TransferError::MissingRingConfig)?
+            .auditor_pubkey;
+        let mut staged = self.stage(auditor_pk)?;
+        let spend_inputs = RingSpendInputs {
+            indexer: environment.indexer,
+            tree: staged.tree,
+            spends: &staged.proof_inputs.input_utxos,
+        }
```
Atamanov, line 261: Consider reading and validating the tree policy before requesting inclusion and non-inclusion proofs

custom-rings/sdk/src/transfer.rs
```diff
@@ -465,15 +597,20 @@ fn frame_dummy_outputs(proof_inputs: &mut SppProofInputs) -> Result<(), Transfer
     Ok(())
 }
 
+/// Inclusion hashes and nullifiers for one spend set.
+type SpendQueries = (Vec<[u8; 32]>, Vec<[u8; 32]>);
```
Atamanov: Please consider usage of a named context and query structures so these values cannot be exchanged accidentally

custom-rings/sdk/src/transfer.rs
```diff
@@ -183,17 +197,85 @@ impl<'a> CustomRingTransfer<'a> {
...
+            indexer: environment.indexer,
+            tree: staged.tree,
+            spends: &staged.proof_inputs.input_utxos,
+        }
+        .load_async()
+        .await?;
+        let allow_dummy_inputs =
+            read_dummy_input_policy_async(environment.rpc, staged.tree).await?;
+        let ring_result = staged.ring_proof(spend_inputs, allow_dummy_inputs)?;
+        let spp_proof = ProofCompressed::try_from(
```
Atamanov: This serializes two independent prover requests. Once ring_result exists, the custom-ring request depends only on private_tx_hash. Prove does not need to wait for prove_transfer_ring. Please consider constructing the custom-ring request first and joining both prover futures, or documenting why prover serialization is required.

Atamanov (commented):
prove / prove_async copy the same protocol; Option plus a comment is what keeps the order. Consider consuming each step into the next type so a skipped step does not compile.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: Done in a6a60925. The lifecycle suite's second hop (`auditor_sees_every_ring_transfer`) is now proven over both transports and lands the async one, so `prove_async` runs end to end against the live chain, Photon and prover; `AsyncHopParity` in custom-rings/test/tests/ring.rs holds it to the blocking proof of the same note. Everything the spent note fixes -- circuit, nullifiers, transaction viewing key, owner signers, expiry -- is asserted equal, and `private_tx_hash` is asserted different, since each proof draws its own blindings, salt and auditor ciphertext. Root indices are left out on purpose: they record the tree state each read saw, not anything about the transport. The unit test stays as the compile-time `Send` check and now points at the lifecycle test.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: Agreed, and it is gone. `MissingRingConfig` was standing in for "this builder was called twice", which is not what it means. The typestate change below removes the code path entirely: there is no `Option` left to fail on.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: Done in a6a60925, as suggested. `StagedTransfer::witness` consumes the staged transfer and returns `(CustomRingProofRequest, WitnessedTransfer)`; `WitnessedTransfer` holds no `pending_proof` and is the only type `finish` is defined on. So skipping a step or repeating one does not compile, nothing is checked at run time, and the unrelated `MissingRingConfig` is gone.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: Done in a6a60925, in both `prove` and `prove_async`. `read_dummy_input_policy` now runs before `RingSpendInputs::load`, so a tree that is absent, owned by another program, or not a tree account fails before Photon has served a full inclusion and non-inclusion proof set that nothing can use.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: Done in a6a60925. `SpendQueries` is a struct with named `utxo_hashes` and `nullifiers` fields, and both `load` and `load_async` destructure it by name. Both halves were `Vec<[u8; 32]>`, so handing the nullifiers to the inclusion query type-checked before and does not now.

custom-rings/sdk/src/transfer.rs
sergeytimoshin: You are right that the dependency was only on `private_tx_hash`, and a6a60925 fixes it. `StagedTransfer::witness` now returns the custom-ring request alongside the SPP witness, and `prove_async` issues both prover calls under `futures::future::try_join` instead of awaiting one before building the other. One caveat worth recording: whether they actually prove at once is the prover's call, not ours -- its sync admission control bounds in-request proving and one gnark proof already spreads across every free core, so against a single instance the server may still serialize them. That bound belongs there rather than in a caller that cannot see the fleet, and the client no longer imposes an ordering the protocol never required. The blocking path is unchanged; it has no way to express this.

sergeytimoshin:
Addressed in a6a60925. Taking the review body first: the stages are types now, exactly as suggested. `StagedTransfer::witness` consumes the staged transfer and is the only way to reach `WitnessedTransfer`, which is the only type `finish` is defined on, so a skipped or repeated step does not compile. The `Option`, the comment that was holding the order, and the unrelated `MissingRingConfig` are all gone.

The rest, one per thread:

- The two prover calls no longer serialize. `witness` hands back the custom-ring request alongside the SPP witness, and `prove_async` issues both under `try_join`. Whether they then prove at once is the prover's admission control to decide, not a caller's.
- Both paths read and validate the tree before asking Photon for inclusion and non-inclusion proofs.
- `SpendQueries` is a struct with named fields rather than a pair of same-typed `Vec<[u8; 32]>`.
- `prove_async` now runs end to end in the lifecycle suite. The second hop of `auditor_sees_every_ring_transfer` is proven over both transports and lands the async one, with the blocking proof of the same note as the parity reference.

`cargo fmt`, workspace `clippy -D warnings`, `cargo machete` and `cargo test -p custom-ring-sdk` are clean locally; the lifecycle suite runs in CI.

Atamanov (approved):
Same remark regarding the comments

PR #275, opened 2026-08-27
ts: list the rings registered with the shielded pool

sergeytimoshin:
There is no way to discover which rings exist. A ring's own config lives under its own program, so that side cannot be enumerated: each ring is a separate program, and you would need its address already to look it up.

The pool keeps a `RingConfig` per ring, and the pool is a single known program. That side is the directory.

## Change

`listRegisteredRings(rpc)` returns every registered ring: config address, ring program, authority, enable flag, paused. Filtered at the RPC on size and discriminator, because the pool also holds trees and asset registries and an unfiltered scan would download all of them to find a handful.

A paused ring is still listed. Hiding it would leave a wallet holding balance there unable to see where it is; a caller choosing a deposit target filters deliberately.

## Bug it uncovered

`decodeRingConfig` could not read a single real account.

`RingConfig` is 68 bytes and the program asserts it:

```rust
pub struct RingConfig {
    pub discriminator: u8,
    pub authority: Address,
    pub program_id: Address,
    pub ring_authority_transact_is_enabled: u8,
    pub paused: u8,
    pub bump: u8,
}
const _: () = assert!(RingConfig::SIZE == 68);
```

The decoder expected **67** and read `bump` out of the `paused` byte. `decodeAccount` rejects a length mismatch outright, so it threw on every account. `RingConfigAccount` had no `paused` field to put it in either.

Devnet currently holds 26 of these, all 68 bytes:

```
$ getProgramAccounts sppXZU5… filter discriminator=4
26 accounts, len=68 each
CFXxaVUTKtHr4yrL7zxWbeP9E2dcB15a7cwuVG1hGHP
  discriminator 4  enabled 0  paused 0  bump 255
```

Reading `bump` at the paused offset would have returned 0 rather than 255, so the field was wrong as well as the length.

## Tests

The account is built byte for byte as the program lays it out, and one test sets `paused` and `bump` to different values so reading either for the other fails. Plus: the query is filtered to ring configs, a listed ring reports the program a utxo names, and an unreadable record makes the whole call throw rather than shortening the list.

That last one changed after review. A wallet could not tell "no rings" from "the registry did not answer", because an unsupported RPC method became an empty array and an undecodable record was skipped. Both now raise a typed `ClientError`. The skip was the worse of the two: the query already filters on `dataSize: 68` and the ring-config discriminator, so a record that does not decode can only be one the RPC returned in spite of the filters -- and then the records that did decode are no more trustworthy than the one that did not. An empty array now means exactly that the pool has no rings registered.

360 tests pass; typecheck and lint clean.

## Why

We are adding a ring picker to a wallet UI: list the deployed rings, choose one, deposit into it.


sdk-libs/ts/src/ring/registry.ts
```diff
@@ -0,0 +1,103 @@
...
+                ) as Base64EncodedBytes,
+              },
+            },
+          ],
+        })
+        .send({ abortSignal }),
+    );
+  } catch (error) {
+    if (error instanceof ClientError && error.code === "CLIENT_UNSUPPORTED_RPC_METHOD") {
+      return [];
```
Atamanov: A wallet cannot distinguish “no rings” from “registry unavailable or incomplete". Please consider propagating typed errors, or returning an explicit completeness/partial-result state.

sdk-libs/ts/test/ring-registry.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    numeric = numeric * 58n + BigInt(digit);
+  }
+  const body: number[] = [];
+  while (numeric > 0n) {
+    body.unshift(Number(numeric & 0xffn));
+    numeric >>= 8n;
+  }
+  const leadingZeroes = value.length - value.replace(/^1+/, "").length;
+  return Uint8Array.from([...Array.from<number>({ length: leadingZeroes }).fill(0), ...body]);
+}
```
Atamanov: Please consider importing getAddressEncoder from "@solana/kit"

sdk-libs/ts/src/ring/registry.ts
sergeytimoshin: Fixed in 461336c6 by propagating typed errors, so the list is now complete or the call throws.

- An RPC that refuses `getProgramAccounts` no longer returns `[]`; the `CLIENT_UNSUPPORTED_RPC_METHOD` `ClientError` propagates.
- A returned account that does not decode no longer gets skipped; it raises `CLIENT_INVALID_RPC_RESPONSE` with the offending index, expected size and actual size. The query already asks for 68-byte accounts carrying the ring config discriminator, so a record that is not one means the response did not honour the filters, and the accounts that did decode are no more trustworthy than the one that did not. Skipping it reported a short list as a complete one.

An empty array now means the pool has no rings registered and nothing else. The signature and the RPC-side filters are unchanged, so a caller that would rather degrade than fail catches the `ClientError` and decides for itself what an unreadable registry should show. Tests cover both refusals.

sdk-libs/ts/test/ring-registry.test.ts
sergeytimoshin: Done in 461336c6. The hand-written base58 decoder and its alphabet are gone; the fixture now builds the account with `getAddressEncoder()` from `@solana/kit`.

Atamanov (approved):

PR #276, opened 2026-08-27
wallet: leave ring-bound utxos to the ring spend path

sergeytimoshin:
A default-ring transfer will select a ring-bound utxo and fail at the prover.

`select_inputs` filters on spent, asset and tree, but not on the ring binding:

```rust
for entry in wallet.utxos.iter().filter(|entry| {
    !entry.spent && entry.utxo.asset == asset && entry.output_context.tree == tree
}) {
```

A ring-bound utxo's commitment covers its ring; the default-ring circuit does not. So the witness is built over an input the circuit cannot account for, the prover refuses it, and the caller sees a prover failure with nothing pointing at the input that caused it.

Split and merge already refuse one, through `is_plain_utxo`. Transfer was the gap.

## Change

Adds `ring_program_id.is_none()` to the filter. A ring balance then looks unavailable to the default path, which is accurate -- it is spendable only through the ring's own path -- and the caller gets `InsufficientBalance` naming what is actually reachable instead of an opaque prover error.

## Test

Pins both halves: a plain utxo still covers its amount, and a ring-bound one is not reachable even when it alone would cover the request.

`cargo test -p zolana-wallet` is green (105 + 12); fmt and clippy clean.

## How it turned up

Depositing into a custom ring from a wallet UI, then trying an ordinary private transfer. The deposit worked, the transfer came back `ExternalProver`, and the cause was several layers from the message.

### DIFF
--- a/cli/src/wallet_cli/transaction.rs
+++ b/cli/src/wallet_cli/transaction.rs
@@ -3,7 +3,10 @@ use solana_signer::Signer;
 use zolana_client::{Rpc, RpcSendTransactionConfig, SolanaRpc, ZolanaClient};
 use zolana_transaction::Address;
 use zolana_wallet::{
-    actions::{submit::MergeMaterial, transaction::is_plain_utxo},
+    actions::{
+        submit::MergeMaterial,
+        transaction::{is_default_ring_spendable, is_plain_utxo},
+    },
     create_merge, create_split, create_transfer_sync, sign_private_transaction_sync,
     submit_merge_transaction, MergeParams, SplitParams, SubmitMergeTransaction, TransferParams,
 };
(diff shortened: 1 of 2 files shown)


Atamanov (commented):
The new filter works, but tree resolution still includes ring-bound UTXOs. A plain UTXO on T1 plus a ring UTXO on T2 fails with AmbiguousTree before selection. Could we conside one default-spendable predicate for both tree resolution and input selection across transfers and withdrawals, and add a two-tree public-path tests?

sergeytimoshin:
@Atamanov good catch, that was a real hole: selection skipped the ring-bound utxo but tree resolution still counted it, so T1-plain + T2-ring failed with `AmbiguousTree` before selection ever ran. Fixed in 6b50e5b.

**One predicate.** `is_default_ring_spendable(entry) -> entry.utxo.ring_program_id.is_none()` now names the rule once and both halves of a default-ring spend use it:

- `create_transfer_with_recipient`'s `resolve_spend_tree`
- `select_withdrawal_inputs`'s two `resolve_spend_tree` calls (and the utxo it names in `InputUtxoTreeMismatch`, so the reported hash is one the spend could actually have taken)
- `select_inputs`

The three `|_| true` predicates are gone; every `resolve_spend_tree` caller now passes the predicate its own selection applies. `is_plain_utxo` is defined in terms of it, so split and merge inherit the same definition of a ring binding, and the CLI's `utxos` listing classifies `ring` with it rather than its own `ring_program_id.is_some()`.

**Two-tree tests**, plain 10 on the default tree plus ring-bound 100 on another:

- `create_withdrawal_spends_the_plain_tree_when_a_ring_balance_sits_on_another` -- the public path directly
- `create_transfer_sync_public_withdrawal_ignores_a_ring_balance_on_another_tree` -- the public path through the unregistered-recipient fallback, which is where the pre-branch `resolve_spend_tree` bit
- `create_transfer_sync_ignores_a_ring_balance_on_another_tree` -- the private path, same layout
- `resolve_spend_tree_ignores_ring_bound_utxos_on_other_trees` -- the resolution itself
- `create_withdrawal_reports_no_balance_when_all_of_it_is_ring_bound` -- an all-ring wallet reports no balance instead of resolving a tree selection then refuses

The first three fail on the previous commit (`AmbiguousTree`), so they pin the fix rather than the current behaviour.

`cargo test -p zolana-wallet` 110 + 12 green, `cargo test -p zolana-cli` 46 green, `cargo check --workspace --all-targets --features zolana-client/proofs` clean, fmt and clippy clean.

Atamanov (dismissed):
Comments are way too long for a one-line predicate and repeat the same explanation several times. Worth cleaning up

ananas-block (approved):

PR #277, opened 2026-08-27
refactor: batched tree to replace bloom filter

ananas-block:
**Changes:**
1. nullifier tree
    1. removed bloom filter, instead we store nullifiers in individual pda accounts
    2. logic to zero out roots is not necessary anymore because root history array length is equal to the number of zkp batches
    3. 
2. shielded pool program
    1. tree accounts are PDAs now
    2. creates 1 PDA account, and derives the canonical bump
    3. 
4. removed bloom filter crate - it is not necessary anymore
5. zolana-event
    1. `BatchAddressAppendEvent` -> `NullifierTreeUpdateEvent`
    2. `EventKind::BatchAddressAppend` -> `EventKind::NullifierTreeUpdate`
    3. 

Notes:
1. Spec changes are too verbose.
2. CLOSE_NULLIFIER_PDAS should be permissioned to a new authority, that should be configured the same as foresters at init
3.  SET_TREE_FEES should have a new authority as well
4. do we check that the next tree that is being created has the correct tree id?
5.


program-libs/interface/src/state/tree.rs
```diff
@@ -2,14 +2,14 @@ use zolana_tree::{InitAddressTreeAccountsInstructionData, TreeAccount};
...
-// Production batched-address-tree parameters. The bloom filter size and hash
-// count are now compile-time const generics of the tree layout.
+// Production batched-address-tree parameters.
 pub const ADDRESS_TREE_INPUT_QUEUE_BATCH_SIZE: u64 = 30_000;
 pub const ADDRESS_TREE_INPUT_QUEUE_ZKP_BATCH_SIZE: u64 = 250;
 pub const ADDRESS_TREE_HEIGHT: u32 = 40;
 pub const ADDRESS_TREE_ROOT_HISTORY_CAPACITY: u32 = 120;
 /// Lamports reimbursed for each applied nullifier-tree ZKP batch.
 pub const FORESTER_REIMBURSEMENT_LAMPORTS: u64 = 5_000;
+pub const TREE_WORKING_CAPITAL_LAMPORTS: u64 = 10_000_000_000;
```
ananas-block: why?

programs/shielded-pool/Cargo.toml
```diff
@@ -28,6 +28,7 @@ groth16-solana = { workspace = true }
 light-array-map = { workspace = true }
 solana-loader-v3-interface = { workspace = true, features = ["serde"] }
 zolana-account-checks = { workspace = true }
+zolana-batched-merkle-tree = { workspace = true }
```
ananas-block: we should take a reexport from the tree crate

program-libs/batched-merkle-tree/src/queue.rs
```diff
@@ -1,98 +1,50 @@
...
-        bloom_filters,
-        hash_chain_store.data,
-        batch_index,
-    )?;
-    // Keep the bounded hash-chain length header consistent with upstream:
-    // length == number of hash-chain slots written so far.
+    current_batch.add_to_hash_chain(value, hash_chain_store.data)?;
     *hash_chain_store.length = current_batch.get_hash_chain_store_len();
 
-    // 3. If batch is full, increment currently_processing_batch_index.
```
ananas-block, line 77: restore the step comments

sdk-tests/zk-program-swap/sdk/src/instructions/make/instruction.rs
```diff
@@ -48,21 +51,26 @@ impl Make {
         if let Some(marker) = spp_proof.messages.first_mut() {
             marker.data = Vec::new();
         }
+        let nullifier_markers = nullifier_marker_accounts(
```
ananas-block: rename nullifier_markers to nullifier_pdas everywhere

sdk-tests/dynamic-swap/test/tests/shared.rs
ananas-block, line 1: this diff seems unnecessary

program-libs/batched-merkle-tree/src/queue.rs
ananas-block: Restored, adapted to the bloom-filter-free code: the doc comment with the three-step list is back, plus the inline `// 1.` / `// 2.` / `// 3.` markers and the hash-chain length invariant comment. The rotation rationale that was deleted along with the inlined arithmetic now sits on `QueueBatches::rotation` in `queue_batch_metadata.rs`, which is where that arithmetic moved. Comments that only described bloom-filter behaviour stay deleted.

sdk-tests/zk-program-swap/sdk/src/instructions/make/instruction.rs
ananas-block: Renamed repo-wide: `NullifierMarker` -> `NullifierPda`, `nullifier_marker_accounts` -> `nullifier_pda_accounts`, the `nullifier_markers` account fields and slot labels, `close_nullifier_markers` (instruction, builder, data struct, program module), the three error variants, the forester module and its CLI verb (`close-markers` -> `close-nullifier-pdas`), the TS surface (`nullifierPda` / `nullifierPdaAddress` / `nullifierPdaAccounts` / `closeNullifierPdas`), the invariant IDs (`INV-CLOSE-MARKER-*` -> `INV-CLOSE-PDA-*`), and the prose in docs/spec.md, the bmt spec/README and the invariants files. Files and modules were moved with git mv.

Unchanged on purpose: the PDA seed is still `b"nullifier"` (it never contained "marker"), instruction tag 18, and error codes 7049-7051. The pinned vector test still yields FketprhoGrMJG7tu9XaXEXhm4vCqzEubwMPFm874xtMm / bump 252, and `error_codes_are_stable` passes. Unrelated markers (the swap program`s `MarkerData` / marker messages, the confidential-output marker, `PhantomData`) were left alone.

sdk-tests/dynamic-swap/test/tests/shared.rs
ananas-block: You are right that it was repeated; it was in eight tree-creation sites. The funding change itself is load-bearing though: `NullifierPdaRent::tree_remaining` rejects with 7049 if paying a PDA rent would drop the tree below its own rent minimum, so a tree created with bare rent exemption fails on its first transact. What was unnecessary was doing the arithmetic in every caller.

Collapsed to one helper. `zolana_interface::state::tree_creation_lamports(params, tree_rent, pda_rent)` is now the single formula, and `zolana_program_test::create_tree_account_ix(rpc, payer, tree, params)` is the single rpc wrapper. Each call site is one line:

    let alloc_ix = create_tree_account_ix(&rpc, &payer.pubkey(), &tree.pubkey(), &address_tree_params())?;

`create_tree_instructions` builds on the same path, and cli/src/wallet_cli/tree.rs and xtask/src/create_release.rs now call the interface helper, which also removes an unchecked `rent + working_capital` add in the CLI.

I could not fold this into `system_create_account_ix` itself: that helper is also used for mints, SPL accounts and ring accounts, so it must not know about tree working capital.

I also audited the rest of the branch for incidental churn. `git diff main...HEAD -w --ignore-blank-lines` yields the same file set as the plain diff, so there is no whitespace-only, blank-line-only or import-reordering-only hunk. The two candidates I checked both turned out justified: the `repeat().take()` -> `repeat_n` change in transact/functional.rs is required (clippy::manual_repeat_n fires there and CI runs -D warnings), and the CU_BENCHMARK.md header paragraph just regenerates text that already existed in the generator on main.

program-libs/event/src/tag.rs
```diff
@@ -32,6 +32,9 @@ pub const RING_TRANSACT: u8 = 15;
 pub const RING_MERGE_TRANSACT: u8 = 16;
 pub const RING_AUTHORITY_TRANSACT: u8 = 17;
 
+// Permissionless maintenance.
```
ananas-block, line 35: should be gated to foresters same as batch updates

program-libs/interface/src/instruction/builders/merge_transact.rs
```diff
@@ -29,14 +30,18 @@ impl MergeTransact {
...
-        let accounts = vec![
+        let mut accounts = vec![
             AccountMeta::new(self.input_tree, false),
             AccountMeta::new(self.output_tree, false),
             AccountMeta::new(self.payer, true),
             AccountMeta::new_readonly(self.user_record, false),
             AccountMeta::new_readonly(Pubkey::default(), false),
-            AccountMeta::new_readonly(PROGRAM_ID_PUBKEY, false),
         ];
+        accounts.extend(nullifier_marker_accounts(
```
ananas-block: nullifiers should be after `PROGRAM_ID_PUBKEY`

ananas-block:
Heads-up on a pre-existing test bug found while verifying the review fixes: `just test-shielded-pool` (the proofs tier, not covered by `test-hermetic`) was red on this branch.

`nullifier_pdas_proof.rs` `transact_tops_up_prefunded_nullifier_pdas` compared the whole post-transact tree `Account` against a pre-transact snapshot with only `lamports` patched. A successful transact necessarily mutates the tree data (queue hash chain, next_index, UTXO root), so that assertion could never hold. `7f83969e` ("assert tree lamports, owner, and size after a spend, not its bytes") introduced `assert_tree_lamports_after_spend` for exactly this reason, but this call site kept the byte compare.

Changed it to assert on `(lamports, owner, data.len(), executable)`, matching the helper, keeping the custom expected lamports (`+ forester_fee - (rent - underfunded)`). The close-path compare in `assert_close_nullifier_pdas` is left as a full byte compare on purpose: `close_nullifier_pdas` really does leave the tree data untouched.

`just test-shielded-pool` is now green: 84 + 1 + 323 + 1 + 6, zero failures.

program-libs/interface/src/instruction/builders/merge_ring.rs
```diff
@@ -52,14 +55,18 @@ impl MergeRing {
...
             AccountMeta::new_readonly(ring_config, auth_signer),
             AccountMeta::new(self.payer, true),
             AccountMeta::new_readonly(Pubkey::default(), false),
-            AccountMeta::new_readonly(PROGRAM_ID_PUBKEY, false),
         ];
+        accounts.extend(nullifier_pda_accounts(
+            &self.input_tree,
+            self.data.nullifiers.iter(),
+        ));
+        accounts.push(AccountMeta::new_readonly(PROGRAM_ID_PUBKEY, false));
```
ananas-block: insert nullifier accounts after static acccounts

program-libs/interface/src/instruction/instruction_data/close_nullifier_pdas.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "borsh")]
+use borsh::{BorshDeserialize, BorshSerialize};
+
+#[cfg_attr(feature = "borsh", derive(BorshDeserialize, BorshSerialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CloseNullifierPdasData {
```
ananas-block: why does this instruction need any instruction data at all?

program-libs/interface/src/state/nullifier_pda.rs
```diff
@@ -0,0 +1,18 @@
+#[cfg(feature = "borsh")]
+use borsh::{BorshDeserialize, BorshSerialize};
+
+pub const NULLIFIER_PDA_SEED: &[u8] = b"nullifier";
+pub const NULLIFIER_PDA_SIZE: usize = 9;
+
+#[cfg_attr(feature = "borsh", derive(BorshDeserialize, BorshSerialize))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct NullifierPda {
```
ananas-block, line 9: add tree ID u16, and a counter in the protocol config that tracks the number of trees

program-libs/interface/Cargo.toml
```diff
@@ -21,16 +20,16 @@ solana-address = { workspace = true, features = ["bytemuck"] }
...
 solana-pubkey = { workspace = true }
 groth16-solana = { workspace = true, optional = true }
-zolana-batched-merkle-tree = { workspace = true }
 zolana-hasher = { workspace = true, features = ["sha256"] }
 solana-program-error = { workspace = true }
 thiserror = { workspace = true }
 wincode = { workspace = true }
 zolana-event = { workspace = true }
-zolana-tree = { workspace = true, optional = true }
+zolana-tree = { workspace = true }
```
ananas-block: can we make it optional again?

docs/spec.md
```diff
@@ -1196,7 +1197,8 @@ The single public signal is `public_input_hash`, a Poseidon hash chain over a sh
...
+| Nullifier PDA | One 10-byte program-owned PDA per queued nullifier: `[b"nullifier", tree, nullifier]` storing `{ queue_index: u64, tree_id: u16 }` (Borsh, no discriminator). Created by the inserting instruction and funded from the tree; it makes a second queue insertion of the same pending nullifier fail. Closed permissionlessly by `close_nullifier_pdas` once `queue_index < tree.close_before_index` and `tree_id` equals the tree's, returning its lamports to the tree. See `program-libs/tree/nullifier_tree_spec.md`. |
```
ananas-block, line 1201: too verbose

docs/spec.md
```diff
@@ -1216,20 +1218,35 @@ struct ProtocolConfig {
...
 creation authority.
 
+The on-chain layout is `#[repr(C)]` with a leading `discriminator: u8` and the
+fields in the order `protocol_authority`, `tree_creation_authority`,
+`forester_authority`, `ring_creation_authority`, `fee_authority`, the three
+flags, `next_tree_id` (166 bytes). Adding `fee_authority` was a breaking layout
+change: a protocol config deployed before it must be re-initialized after the
+program upgrade (there is no migration instruction), and trees deployed before
+it decode an all-zero fee schedule (no fees charged, no reimbursements paid)
+until `set_tree_fees` runs on them.
```
ananas-block, line 1245: remove

docs/spec.md
```diff
@@ -1284,16 +1309,17 @@ Usage by instruction:
...
+| create_tree | Tag 2; gated by `protocol_config.tree_creation_authority` unless `tree_creation_is_permissionless`. Accounts: `payer` (signer), `authority` (signer), `protocol_config`, the tree PDA, system program; data `{ tree_id: u16, nullifier_params, fees: TreeFeeSchedule }`. Repeated once per 10 KiB allocation step within one transaction: the first step requires `tree_id == protocol_config.next_tree_id`, increments it, and creates the PDA funded by `payer` with its rent plus the nullifier-PDA working capital; the step that reaches the full size initializes the shared Tree account (nullifier tree + queue, UTXO tree), stores `fees` as submitted (no check against `nullifier_params.input_queue_zkp_batch_size`; payouts are capped by `fee_balance`), and starts `fee_balance` at 0. The default schedule (`default_tree_fees(Z)`) is `append_reimbursement = 5_000`, `close_reimbursement = 170`, `fee_per_nullifier = ceil((5_000 + Z * 170) / Z)`: 190 for `Z = 250`, 670 for `Z = 10`, so insertions pay exactly the maintenance cost and nothing accumulates. |
```
ananas-block, line 1320: too verbose

docs/spec.md
```diff
@@ -1284,16 +1309,17 @@ Usage by instruction:
...
-| batch_update_nullifier_tree | Tag 4; gated by `protocol_config.forester_authority`; inserts queued nullifiers into the nullifier tree via a batch ZKP and emits the batch address-append event. |
+| batch_update_nullifier_tree | Tag 4; gated by `protocol_config.forester_authority`; accounts `authority` (signer), `protocol_config`, `tree` (writable), `reimbursement_recipient` (writable). Inserts queued nullifiers into the nullifier tree via a batch ZKP and emits the batch address-append event. When an applied update marks the previously inserted queue batch reclaimable it zeroes the root-history slots that predate that batch's final root and advances `close_before_index`, releasing that batch's nullifier PDAs for `close_nullifier_pdas`. After applying `num_update` ZKP batches it pays `min(fees.append_reimbursement * num_update, fee_balance)` from the tree to `reimbursement_recipient` and debits `fee_balance` by the paid amount; a shortfall never fails the update, so liveness survives a fee increase. The recipient must not be program-owned (`InvalidReimbursementRecipient`, 7055), checked before any state change. |
```
ananas-block, line 1322: too verbose

program-libs/hasher/Cargo.toml
```diff
@@ -11,23 +11,22 @@ edition = "2021"
 default = ["std"]
 alloc = []
 std = ["alloc"]
-poseidon = ["dep:light-poseidon", "dep:ark-bn254", "dep:ark-ff"]
 sha256 = ["dep:sha2"]
 keccak = ["dep:sha3"]
```
ananas-block, line 15: do we even need the other features if we gate the deps not target os solana?

program-libs/interface/src/instruction/builders/create_tree.rs
```diff
@@ -1,53 +1,50 @@
...
-        params: Option<InitAddressTreeAccountsInstructionData>,
-    ) -> Instruction {
-        let mut data = vec![tag::CREATE_TREE];
-        if let Some(params) = params {
-            params
-                .serialize(&mut data)
-                .expect("shielded-pool instruction serialization is infallible");
-        }
-
+    pub fn step(&self) -> Instruction {
```
ananas-block, line 32: whats a better name?

program-libs/interface/src/state/tree.rs
```diff
@@ -1,30 +1,59 @@
...
+pub const DEFAULT_APPEND_REIMBURSEMENT_LAMPORTS: u64 = 5_000;
+pub const DEFAULT_CLOSE_REIMBURSEMENT_LAMPORTS: u64 = 170;
+
+pub fn default_tree_fees(zkp_batch_size: u64) -> TreeFeeSchedule {
+    TreeFeeSchedule::at_cost(
+        zkp_batch_size,
+        DEFAULT_APPEND_REIMBURSEMENT_LAMPORTS,
+        DEFAULT_CLOSE_REIMBURSEMENT_LAMPORTS,
+    )
+    .unwrap_or_default()
```
ananas-block, line 22: we should include v0 or v1 transaction size as parameter since that is what drives DEFAULT_CLOSE_REIMBURSEMENT_LAMPORTS cost, v0 tx size 1232, v1 tx size 4kb.

program-libs/interface/src/pda.rs
```diff
@@ -106,11 +107,68 @@ pub fn ring_auth_with_bump(ring_program: &Pubkey, bump: u8) -> Result<Pubkey, Pu
...
+
+pub fn tree_with_bump(tree_id: u16) -> (Pubkey, u8) {
+    Pubkey::find_program_address(
+        &[TREE_PDA_SEED, &tree_id.to_le_bytes()],
+        &shielded_pool_program_id(),
+    )
+}
+
 #[cfg(test)]
 mod tests {
```
ananas-block, line 129: can we move these tests to the tests dir?

ananas-block:
superseeded

PR #279, opened 2026-08-28
chore: link explorer urls to orbmarkets

sergeytimoshin:
Point the explorer links the ring CLI and the deploy scripts print at orbmarkets.

- `custom-rings/cli/src/status.rs` builds its address link from `https://orbmarkets.io/address`, and the test pinning the devnet and localnet strings moves with it.
- `tools/deploy-devnet.sh` prints the deployed program address at orbmarkets.
- `tools/cli_smoke.sh` prints transaction links at orbmarkets.

The localnet link keeps the `cluster=custom&customUrl=` query it had. That convention is Solana Explorer's, and whether orbmarkets honours it is unverified, so click one localnet link before merging.

### DIFF
--- a/custom-rings/cli/src/status.rs
+++ b/custom-rings/cli/src/status.rs
@@ -13,7 +13,7 @@ use crate::{
     Context,
 };
 
-const EXPLORER: &str = "https://explorer.solana.com/address";
+const EXPLORER: &str = "https://orbmarkets.io/address";
 
 #[derive(Debug, Error)]
 pub enum StatusError {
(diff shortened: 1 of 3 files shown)


sergeytimoshin:
Closing: Orb supports only `cluster=mainnet-beta|devnet|testnet` (https://www.helius.dev/docs/orb/explore-devnet). The localnet arm here emits `cluster=custom&customUrl=`, which is Solana Explorer's convention and has no Orb equivalent, so that link would be wrong. Needs a per-target split before this is worth reopening.

PR #282, opened 2026-08-31
Support for default <-> custom ring flow, tokens, TS SDK refactoring

Atamanov:
This PR makes custom rings a first-class wallet boundary and reshapes the TypeScript SDK around durable state, explicit authority, and reusable domain contracts.

-  The default pool and custom rings now form a complete value flow with explicit entry, transfer, exit, and withdrawal operations, while balances, selection, and history preserve that boundary.
-  SOL, SPL Token, and Token-2022 follow one asset model across wallet and ring operations, with shared settlement, account creation, registry resolution, and exact amount handling.
-  Wallet sync now behaves as a durable transaction, with staged state, persisted cursors, reservations, and one sealed sync-and-save path that supports restart without routine full-history scans.
- Value movement crosses an explicit security boundary where scoped authority sessions limit key lifetime and intent-bound capabilities ensure the approved operation is the operation proved and compiled.
- Load-bearing SDK code now depends on narrow client ports and shared selection, settlement, compilation, transport, persistence, and error policies, which removes rail-specific duplication and concrete-client coupling.
-  Rust behavior, TypeScript behavior, public exports, documentation, package checks, and flow coverage now describe the same integration contract.


sergeytimoshin:
I'd rather use "UTXO" instead of "note" in code & comments to be more aligned with what we already have

sdk-libs/ts/test/prover-client.test.ts
```diff
@@ -426,8 +422,9 @@ describe("dummy prover inputs", () => {
     } as NonInclusionProof;
 
     const converted = createDummyTransferInput(input, 4n, proof);
-    const utxo = circuitUtxo(converted);
+    const utxo = converted.circuit;
 
+    expect({ ...converted }.circuit).toBe(converted.circuit);
```
sergeytimoshin, line 427: doesn't make sense for me

sergeytimoshin (approved):

PR #283, opened 2026-08-31
photon: add getRegisteredAssets

tilo-14:
## Summary
- Add Photon's `getRegisteredAssets` JSON-RPC method: empty `params`, returns `{ context, assets }` with SOL (`assetId` 1) first, then parsed `SplAssetRegistry` rows.
- Uses filtered `getProgramAccounts` (`dataSize` 48 + discriminator memcmp). The list is protocol-wide; there is no `ringProgramId` field.

## Test plan
- [x] `cargo test -p zolana-indexer-api method_markers_have_canonical_names`
- [x] `cargo test -p zolana-indexer-api get_registered_assets_request`
- [x] `cargo test -p photon-indexer get_spl_asset_registry_accounts`
- [x] `cargo test -p photon-indexer get_registered_assets`
- [x] `cargo test -p photon-indexer registers_only_rings_product_methods`
- [x] `cargo test -p photon-indexer --test integration_tests test_documentation_generation`

### DIFF
--- a/services/photon/src/api/method/rings/get_registered_assets.rs
+++ b/services/photon/src/api/method/rings/get_registered_assets.rs
@@ -1,0 +1,258 @@
+use std::collections::HashSet;
+
+use sea_orm::DatabaseConnection;
+use solana_account::Account;
+use solana_pubkey::Pubkey;
+use zolana_indexer_api::{GetRegisteredAssetsResponse, RegisteredAsset, SerializablePubkey};
+use zolana_interface::{pda, state::SplAssetRegistry};
+
+use crate::api::error::PhotonApiError;
+use crate::common::indexer_context::extract as extract_context;
+use crate::rpc::RpcClient;
+
+const MAX_REGISTERED_ASSET_ACCOUNTS: usize = 4096; // parity with ring-rpc audit.rs MAX_ASSET_REGISTRY_ACCOUNTS
+const SOL_ASSET_ID: u64 = 1;
+const SOL_MINT: [u8; 32] = [0u8; 32];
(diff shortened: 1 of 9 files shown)


services/photon/src/api/method/rings/get_registered_assets.rs
```diff
@@ -0,0 +1,258 @@
...
+
+pub async fn get_registered_assets(
+    conn: &DatabaseConnection,
+    rpc: &RpcClient,
+) -> Result<GetRegisteredAssetsResponse, PhotonApiError> {
+    let context = extract_context(conn).await?;
+    let accounts = rpc
+        .get_spl_asset_registry_accounts(&pda::shielded_pool_program_id())
+        .await
+        .map_err(|error| PhotonApiError::UnexpectedError(format!("RPC error: {error}")))?;
```
sergeytimoshin, line 77: if `get_registered_assets` is just a thin wrapper to rpc call, why do we need this in photon?

tilo-14: having it exposed in context to what it is indexing seems consistent

what is your suggestion?


what sandro said: "seems like a useful thing for an indexer to provide"

sergeytimoshin: Why its useful? If this is public on-chain data, why don't get this data directly?

```
let accounts = rpc
        .get_spl_asset_registry_accounts(&pda::shielded_pool_program_id())
        .await
        .map_err(|error| PhotonApiError::UnexpectedError(format!("RPC error: {error}")))?;
```

PR #285, opened 2026-09-01
test(wallet): show a fresh wallet cannot replay a merge

TarekkMA:
Two failing tests in `sdk-libs/ts/test/wallet-sync.test.ts`. Tests only — no source changes.

## What fails

A merge output carries no ciphertext, so it is reconstructed from the notes it spends rather than decrypted. `buildTagIndex` files those outputs into `recipientSites`, and `processStableTags` walks recipient sites before sender sites. So when a split and the merge that consumes it are decoded in the same batch, the merge is reconstructed before the split has stored its notes, `#reconstructMerge` counts the miss, and nothing revisits the site.

- `reconstructs a merge whose inputs arrive in the same batch` expected `{ storedUtxos: 3, undecryptableCandidates: 0 }`, got `{ storedUtxos: 2, undecryptableCandidates: 1 }`
- `reaches the same state whether the merge replays fresh or incrementally` expected `42n`, got `0n`

Existing merge coverage (`reconstructs a ciphertext-free merge from owned spent inputs`) preloads the inputs, which satisfies the dependency and hides the ordering. The Rust sync path in `sdk-libs/transaction/src/wallet/sync.rs` has the same order.

## Reproduce

```bash
cd sdk-libs/ts
npm install
npm test -- wallet-sync

### DIFF
--- a/sdk-libs/ts/test/wallet-sync.test.ts
+++ b/sdk-libs/ts/test/wallet-sync.test.ts
@@ -7,6 +7,7 @@ import { mergeDummyNullifier, mergeOutputBlinding } from "../src/keypair/merge/i
 import { SHIELDED_POOL_PROGRAM_ID, type Bytes16, type Bytes32 } from "../src/interface/index.js";
 import { StateDiscriminator } from "../src/interface/state.js";
 import {
+  AssetRegistry,
   ConfidentialTransfer,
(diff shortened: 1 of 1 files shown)


TarekkMA:
fixed and tests are included in #286

PR #286, opened 2026-09-01
Fix merge replay dependency order

Atamanov:
Wallet replay now preserves same-sync merges, and the TypeScript SDK uses UTXO terminology consistently.
- TypeScript and Rust resolve merges after ordinary outputs, tests cover fresh and chained replay, and version 3 snapshot keys remain unchanged.


tilo-14 (approved):
makes sense

PR #288, opened 2026-09-02
feat(tree): batched nullifier tree with nullifier PDAs and forester reimbursement

ananas-block:
### Summary
Replace the bloom-filter batched Merkle tree with a nullifier tree that records spent nullifiers as program-derived accounts until the forester appends them to the tree.

- merge the batched merkle tree crate into zolana-tree as nullifier_tree, flatten the metadata, queue, and batch layouts into NullifierTreeLayout, and collapse the const generics to the zkp batch count
- replace bloom filters with nullifier PDAs created inline during a spend and closed by the forester once the batch is appended and reclaimable
- make the shielded-pool tree a program-derived account
- add a configurable forester fee schedule and fee balance to the tree account, collected before nullifier PDA creation and paid out on append
- derive tree working capital from the queue batch size
- gate pinocchio and groth16-solana behind zolana-tree features and drop the poseidon feature from zolana-hasher
- forester: close-nullifiers command, close_before_index in info, and nullifier PDA validation before close
- tests: nullifier PDA lifecycle coverage, v0 lookup-table transfers for oversized ring transactions, regenerated CU benchmark
- docs: specify the batched Merkle tree flows, invariants, and nullifier PDA cleanup

### Details

**Changes:**
1. nullifier tree
    1. removed bloom filter, instead we store nullifiers in individual pda accounts
    2. logic to zero out roots is not necessary anymore because root history array length is equal to the number of zkp batches
    3. adds spec for nullifier tree
2. shielded pool program
    1. tree accounts are PDAs now
    2. creates 1 PDA account, and derives the canonical bump
    3. 
4. removed bloom filter crate - it is not necessary anymore
5. zolana-event
    1. `BatchAddressAppendEvent` -> `NullifierTreeUpdateEvent`
    2. `EventKind::BatchAddressAppend` -> `EventKind::NullifierTreeUpdate`
6. zolana-hasher
    1. add canonical field check
    2. remove hash fn features
    3. switch error codes to 8k number space
    4.


program-libs/interface/src/instruction/builders/transact.rs
```diff
@@ -146,7 +162,7 @@ impl Transact {
 #[cfg(test)]
```
ananas-block, line 162: can these tests be in the tests dir?

sergeytimoshin:
`program-libs/tree/src/nullifier_tree/error.rs:52-90`

`impl From<NullifierTreeError> for u32` (error.rs:52) and `impl From<NullifierTreeError> for ProgramError` (error.rs:86) have no caller anywhere in the repository. The only program-side consumer of `NullifierTreeError` is `nullifier_tree_error()` in programs/shielded-pool/src/instructions/shared.rs:35, which maps explicitly onto `ShieldedPoolError` codes in the 7000 space and never touches the conversion. No test and no client asserts a 14xxx code - the whole numbering scheme, including the two "do not reuse the code" reservations for 14009 and 14015, protects codes nothing can emit. It is carryover from the deleted `zolana-batched-merkle-tree` crate.

Three consequences. It contradicts the project error rule ("All program errors must be defined in the interface crate, including the `From<...> for ProgramError` conversion"; "Error codes live in the 7000 space. Pin every code in the `error_codes_are_stable` test") - `TreeError`, the crate's other error type, correctly has no `ProgramError` impl and is converted in the interface crate. It is a second, unpinned numbering scheme beside a pinned one. And it makes `solana-program-error` a non-optional dependency of `zolana-tree` (Cargo.toml) purely for the dead impl, which contradicts the module doc added in this same change: "Nothing is on by default: deserializing a tree account out of bytes needs neither a Solana runtime nor a proof verifier, and a client that only reads the account should not link one" (src/lib.rs).

Fix: Delete both `From` impls and the code table, and drop `solana-program-error` from `program-libs/tree/Cargo.toml`. If a numeric surface is wanted later, add the variants to `ShieldedPoolError` in the interface crate and pin them in `error_codes_are_stable`.

sergeytimoshin:
### `Batch::num_zkp_batches()` is dead public API and shadows `get_num_zkp_batches()`

`program-libs/tree/src/nullifier_tree/batch.rs:132-134`

`pub fn num_zkp_batches(&self) -> usize` (batch.rs:132) has zero callers in the repository, including tests and benches. The only other `num_zkp_batches` hits are an unrelated field on `forester::info::BatchInfo` (forester/src/info.rs:32), populated from `batch.get_num_zkp_batches()` - a different method at batch.rs:317 returning `batch_size / zkp_batch_size` as `u64`.

So the type exports two near-identically-named public methods that return different things by different means: `num_zkp_batches()` yields `hash_chains.len()` (the const-generic array length, `usize`), `get_num_zkp_batches()` yields the computed quotient (`u64`). They coincide only while `NullifierTreeLayout::validate_configuration` holds (layout.rs:87, which pins `batch_size / zkp_batch_size == ZKP_BATCHES`); a corrupt account that parses can hold a layout where they disagree. The doc comment ("Number of cached update slots, one per ZKP batch") names `cached_tree_updates` while the body reads `hash_chains` - same length, different array. A caller reaching for "the number of zkp batches" on this new public surface can easily pick the one nothing uses.

Fix: Delete `num_zkp_batches()`. If the array length is needed internally later, make it private and name it for what it returns.

sergeytimoshin (approved):

PR #289, opened 2026-09-02
ts-sdk: add remote wallet key interface

sergeytimoshin:
Allow TypeScript wallet sync and transaction building to use privacy keys held by a remote service or enclave. The key holder performs decryption, derivation, and proving without returning long-lived secrets to the app.

- Replace `WalletAuthority` with `WalletKeys`, combining batched `ShieldedKeys` operations and `ProofAuthority` methods.
- Provide `LocalKeys` and `LocalShieldedKeys` for in-process keys; callers migrate from `authority` to `keys`.
- Keep transaction approval as a separate optional callback and Solana transaction signing with the app's signer.

Validation: SDK checks and package dry-run pass. All 31 CI checks pass, including TypeScript E2E and localnet suites.


ananas-block:
**Dead code: the decrypt helpers sync no longer calls**

`decryptAnonymous` (`codecs.ts:947`), the `decryptSplit` alias (`:960`) and
`decryptConfidential` (`:1002`) are unreachable after this branch.

On `main`, `sync.ts` was their only `src/` caller:

```
sdk-libs/ts/src/transaction/wallet/sync.ts:751  decryptConfidential(key, txViewingPublicKey, body, salt, site.slot)
sdk-libs/ts/src/transaction/wallet/sync.ts:955  return decryptAnonymous(key, txViewingPublicKey, body, salt, slotIndex)
```

Both call sites are now `#decryptSlot` -> `KeyMemo.decrypt`, and there is no other
`src/` reference. They are also not reachable from outside the package: they are
exported only from `src/transaction/serialization/index.ts`, which is absent from the
`exports` map, and `src/transaction/index.ts` re-exports named lists from it that leave
all three out. So neither internal nor consumer code can reach them; only
`transaction/test/serialization.test.ts` keeps them compiled.

Deleting them is consistent with the design rather than just tidying, because each has
a successor. The old helpers did decrypt-and-decode in one call while holding a
`ViewingKey`; this branch split that into "decrypt through the `ShieldedKeys` batch" +
"decode with a `decode*` function", which is the seam a remote holder needs:

| Removed | Replaced by |
|---|---|
| `decryptAnonymous` | `ShieldedKeys.decrypt({ label: "transfer" })` -> `LocalShieldedKeys.decryptOne` -> `ViewingKey.decryptUtxo`; in sync, `SyncPass.#decryptSlot` (`sync.ts:992`) |
| `decryptSplit` | same, then `decodeSplitBundle(decrypted)` (`sync.ts:856`) |
| `decryptConfidential` | `splitEmbeddedKey(body).rest` + `#decryptSlot` + `decodeConfidential(decrypted)` (`sync.ts:753-762`) |

`splitEmbeddedKey` was made public in this branch (`codecs.ts:964`) precisely to serve
that split.

Suggested cleanup:

- delete the three definitions in `serialization/codecs.ts` and their entries in
  `serialization/index.ts:12,20,22`
- in `transaction/test/serialization.test.ts`, drop the cases that exist only to cover
  them (`:118`, `:139`, `:181`, `:200`, `:448`) and remove the two `decryptAnonymous` /
  `decryptSplit` entries from the destroyed-key list at `:380-381`, keeping the three
  encrypt cases so the `TRANSACTION_KEYPAIR` conversion stays under test

Two carve-outs, both intentional as far as I can tell:

- `decryptConfidentialAsSender` stays. It is live in `sync.ts:595` and
  `ring/audit.ts:253`, and it uses the ephemeral transaction key the caller already
  holds rather than the long-lived viewing key, so it needs no holder round trip.
  `inTransactionCategory` therefore keeps callers (it and `encryptAnonymous`) and
  survives the deletion.
- `decryptRingDepositUtxo` is published API via `./ring` (`src/ring/index.ts:18` ->
  `src/ring.ts`), so removing it would be a breaking change, not a cleanup. It is
  internally unused now, superseded by `ringDepositUtxo` (`ring-deposit.ts:113`) plus
  a `label: "ringDeposit"` decrypt. Keeping it for consumers who decrypt locally is
  reasonable; if you would rather not carry both shapes, this alpha is the moment,
  since the Breaking section is already long.

**One separate question the deletion surfaces.** `inTransactionCategory` converted a
cipher `KeypairError` into `TRANSACTION_KEYPAIR`, and its comment says why: "without it
a malformed published slot escapes as a `KeypairError` no transaction caller catches."
The new decrypt path does not do that conversion -- no `inTransactionCategory`,
`KeypairError` or `TRANSACTION_KEYPAIR` handling in `keys.ts`, `key-memo.ts` or
`sync.ts`. The blast radius changed with it: the decrypt used to run inside each slot's
`catch (error) { this.#recordUndecryptable(error, siteKey); return; }`, so one bad slot
was counted and the sync continued; it now runs in `KeyMemo.resolve()`, outside any
per-slot handler, and `fulfilled()` rethrows the raw cause, so one bad slot fails the
whole sync.

Reachability today is low and I would not call it a live bug: `P256PublicKey.fromBytes`
validates the point at construction, `checkDecryptRequest` validates salt width and slot
range before the cipher, and AES-CTR does not authenticate, so the common case -- a
tag-collision miss under the wrong key -- yields garbage that the decoder rejects inside
the existing per-slot `catch` and stays contained. What is no longer contained is a
genuine cipher error (destroyed key, HKDF failure). If the per-slot containment was
intentional, wrapping the `keys.decrypt` call in `KeyMemo.resolve` restores it.

ananas-block:
**Duplicated protocol math around the dummy slot**

Three spots where the dummy slot's derivation gained a second implementation. No
behavior change; I traced each.

**1. `dummyNullifier` re-spells the canonical derivation.**

`utxo.ts:382` computes `poseidon([hash, utxo.blinding, ZERO_32])` inline. The canonical
form is `NullifierKey.nullifier` (`keypair/nullifier-key.ts:29`),
`poseidon([hash, blind, rightAlign(secret)])`, mirrored in Rust at
`sdk-libs/keypair/src/nullifier_key.rs:40`. The values agree, since `rightAlign` of 31
zero bytes is `ZERO_32`. It is a second place spelling out the Poseidon input order, and
it is TS-only: Rust's `SppProofInputUtxo::new_dummy()` (`instructions/types.rs:40`) holds
`NullifierKey::from_secret([0u8; BLINDING_LEN])` and derives through the canonical
method, as TS did before this branch.

The branch made `utxo.ts` import `NullifierKey` as a type so `ProofInputUtxo` holds no
key material; a value import would undo that. Exporting
`zeroKeyNullifier(utxoHash, blinding)` from `keypair/nullifier-key.ts` keeps both that
property and one spelling of the Poseidon call.

`dummyNullifier` also passes `asset` and `amount` to `fullOwnerUtxoHash`, whose `dummy`
branch (`utxo.ts:133-144`) reads only `blinding`.

**2. The canonical-dummy predicate changed meaning.**

```diff
-  if (!isZeroNullifierKey(input.nullifierKey)) return "nullifier_key";
+  if (!isZero(input.nullifierPublicKey)) return "nullifier_public_key";
```

A zero secret hashes to `poseidon([0])`, not to zero, so these test different things.
`poseidon([0])` is what `ProofInputUtxo.dummy()` produced before, and what Rust's dummy
still holds. Safe either way: a dummy's `nullifierPublicKey` reaches neither the proof
nor the instruction data. `inputCircuitUtxo` returns `owner = 0n` for `dummy = true`
(`assembly.ts:430`), `createDummyInput` writes `ownerPublicKeyHash: asField(0n)`
(`:403`), and `realInputContexts` filters dummies at `builders.ts:124`. The field is now
a sentinel holding a new value, which the obvious reading of it as a hash misses. Worth
a comment.

**3. Four copies of the secret-copy idiom.**

`copyViewingKey` / `copyNullifierKey` (`transaction/wallet/keys.ts:341,350`) repeat
`ShieldedKeypair.viewingKey()` / `nullifierKey()` (`keypair/shielded.ts:270,279`) line
for line: read `secretBytes()`, rebuild, wipe. One helper beside the key classes would
put the wipe-on-throw discipline in a single place.

ananas-block:
**Naming: `TRANSACTION_WALLET_AUTHORITY_MISMATCH` outlives the model it names**

Two throw sites, both added by this branch: `LocalShieldedKeys.#checkIdentity`
(`keys.ts:286`) and `checkKeysIdentity` (`:303`). Both mean "these keys are not this
wallet's". No wallet-authority type is left in `sdk-libs/ts`.

`CHANGELOG.md:14` removes `WalletAuthority`, `SpendAuthority`, `SyncWalletAuthority` and
the rest. Line 33 of the same Breaking section tells anyone passing the new `keys` that a
mismatch fails with `TRANSACTION_WALLET_AUTHORITY_MISMATCH`. A reader following that
entry looks up a type the entry above deleted.

Nine occurrences in total: the two throws, the `TRANSACTION_ERROR_CODES` entry, the
changelog line, and five across three test files. Something like
`TRANSACTION_KEYS_IDENTITY_MISMATCH` would say what the two checks actually compare.
Cheap to do while the alpha's surface is already breaking, and it gets more expensive
once someone matches on the code.

ananas-block (dismissed):

sdk-libs/ts/src/transaction/wallet/key-memo.ts
```diff
@@ -0,0 +1,193 @@
...
+      derived.forEach((value, index) => {
+        const request = derives[index];
+        if (request !== undefined) this.#derived.set(request[0], value);
+      });
+      keys.forEach((viewingKey, index) => {
+        const request = transactionKeys[index];
+        if (request !== undefined) this.#transactionKeys.set(request[0], viewingKey);
+      });
+    } catch (cause) {
+      for (const key of minted) key.destroy();
```
ananas-block: `minted` is whatever the holder returned, not necessarily an array. When `transactionKeys` answers with a non-array, `checkedAnswers` raises `TRANSACTION_KEYS_BATCH_MISMATCH` and this line throws over it:

```
THROWN: TypeError  TypeError: minted is not iterable
```

`decrypt` and `derive` report that malformation correctly, so this is the one hole in the boundary validation the change exists for. An `Array.isArray(minted)` guard before the loop, skipping non-`ViewingKey` entries, restores the intended error.

sdk-libs/ts/src/transaction/wallet/sync.ts
```diff
@@ -973,81 +1016,119 @@ class SyncPass {
...
+      });
+      pass.processStableTags({ identityTag, index });
+      pass.resolveMergeSites(mergeSites);
+      if (memo.pending()) {
+        await memo.resolve(input.context);
+        continue;
+      }
+      return commitPass(input, pass, index, viewingKeyHistory);
+    }
+    throw new TransactionError("TRANSACTION_KEYS_UNRESOLVED");
```
ananas-block, line 1115: Same code as `transaction/utxo.ts:373`, where it means "a real input cannot be rebound to another tree". Here it means "the holder never converged inside the round budget".

Only the first has a test (`transaction/test/core.test.ts:252`), leaving the budget path, which is the one a remote holder can reach, uncovered. Worth splitting the code and testing it with a holder that answers correctly but never lets the memo cache.

sdk-libs/ts/src/client/prover/merge.ts
```diff
@@ -343,27 +329,23 @@ function checkNullifierRoot(
   }
 }
 
-function validateMergeMaterial(prepared: PreparedMerge, material: MergeMaterialInput): void {
+function validatePreparedMerge(prepared: PreparedMerge): void {
```
ananas-block, line 332: `validateMergeMaterial` checked the caller's `signingPublicKey` and `nullifierKey` against the prepared merge. This checks `PreparedMerge` against itself.

`client.proveMerge({ prepared, keys })` with a `prepared` for one wallet and `keys` for another now assembles fine, `LocalKeys.proveMerge` declines to fill the secret (`client/keys.ts:97`), and it lands as a generic `CLIENT_MISSING_NULLIFIER_SECRET` from the request encoder. `PreparedMerge` is public API, so this is reachable.

Two things to settle: `CLIENT_MERGE_SIGNING_KEY_MISMATCH` now claims "this input's owner is not the prepared signing key" under the old name, and neither it nor `CLIENT_MERGE_NULLIFIER_KEY_MISMATCH` has a test. Checking `keys` against `prepared.nullifierPublicKey` inside `proveMerge` would name the actual problem.

sdk-libs/ts/src/wallet/merge.ts
```diff
@@ -77,27 +91,39 @@ export function createMerge(params: MergeParams): CreatedMerge {
...
+        ...dummySlots.map((slotIndex) => ({
+          kind: "mergeDummyNullifier" as const,
+          firstNullifier,
+          slotIndex,
+        })),
+      ],
+      context,
+    );
+    if (
+      outputBlinding === undefined ||
```
ananas-block: Checks `undefined` and the count. `KeyMemo.checkedAnswers` checks `instanceof Uint8Array` and length 32 for the same class of answer.

A holder returning a string or 16 bytes lands on a `checked<Bytes32>` deserialize error from inside `PreparedMerge` rather than `WALLET_KEYS_BATCH_MISMATCH`.

sdk-libs/ts/src/client/ports.ts
```diff
@@ -102,15 +104,45 @@ export interface ProofReader {
...
+  proveMerge(inputs: MergeInputs, context?: RequestContext): Promise<Proof>;
+}
+
+/**
+ * The one capability that consumes the nullifier secret: completes proof
+ * inputs whose owner it holds the secret for, and proves them. Inputs arrive
+ * with `nullifierSecret` absent on the wallet's own real inputs; dummy slots
+ * carry zero and any other owner's inputs arrive already complete.
+ */
+export interface ProofAuthority {
```
ananas-block, line 123: `ProofAuthority` and `ProofService` declare the same two members, so TypeScript cannot separate them. `LocalKeys.fromKeypair(kp, someWalletKeys)` type-checks and nests an authority inside an authority. It works, since `completeInput` short-circuits on an already-present secret, but nobody meant it.

`checkProofService` and `checkProofAuthority` both reduce to `hasProofMethods`. Give one a nominal marker, or collapse them into one interface.

ananas-block (commented):
Typecheck, lint, format and 858 unit tests pass. The `ShieldedKeys` batch interface and the re-run-the-pass-until-resolved `KeyMemo` loop hang together, and the depth-9 merge chain plus the short/holed-batch tests cover the new machinery.

Five inline comments, plus two top-level ones for findings that fall outside the diff. The `bench/loadtest.ts` breakage and the `KeyMemo.resolve` cleanup are the two I would fix before merge.

ananas-block:
**`sdk-libs/ts/bench/loadtest.ts:43`** (outside the diff, so no inline anchor)

`KeypairWalletAuthority` no longer exists, so this import fails at ESM link time and `npm run loadtest` is dead. Lines 98, 161, 214 and 225 still pass `authority:` as well.

`bench` sits in neither the tsconfig `include` nor the oxlint globs, which is why `npm run check` stayed green. Porting to `LocalKeys`/`keys:` fixes it now; adding `bench` to the typecheck include stops it recurring.

ananas-block:
Smaller ones, none blocking:

- `client/prover/client.ts:372` — `mergeProverRequestBody` has no caller and no test; `proverRequestBody` has both.
- `wallet/private-transaction.ts:132` — the nullifier comes from the entry, derived under `outputContext.tree`, while the `ProofInputUtxo` is built at `DEFAULT_TREE_ID`. One live tree today, but the two used to share a `treeId` and no longer do. `WalletUtxo` carries `tree` as an `Address`, so a comment may be the only cheap fix.
- `transaction/utxo.ts:431` — `dummyNullifier` passes `asset` and `amount` into a branch that ignores both.
- `transaction/wallet/intent.ts:67` — `ownerSolanaAccount` is used by the ring and merge builders; `private-transaction.ts:186,228` still calls `address.solanaAddress()`.
- `instructions/builders.ts:130` — hands back the caller's arrays; `:134` returns a copy.
- `ring/index.ts:23` — `decryptRingDepositUtxo` lost its internal caller to `ringDepositUtxo`. Deliberate public surface, or leftover?
- `wallet/internal.ts:70` — the base64 rewrite is a megabyte-snapshot fix riding along with a keys refactor.
- `snapshot-cipher.ts:56` — `walletSnapshotKey` zeroes its caller's argument, including on the length failure. Naming the parameter `consumeSecret` would say so.

ananas-block (dismissed):

ananas-block (approved):

PR #292, opened 2026-09-03
perf: enable 36 in out utxos

ananas-block:


ananas-block:
https://github.com/helius-labs/zolana/pull/297

PR #295, opened 2026-09-04
prover: fetch a sourced key like any other and log why a custom ring proof failed

sergeytimoshin:
Lets a prover with auto-download serve custom ring proofs from the object store, and gives operators the cause of a custom ring failure.

- EnsureProvingKey no longer refuses a key whose lockfile entry carries a source, so custom_ring.key is fetched and verified against its pinned sha256 like every other key.
- The test that pinned the refusal now asserts a sourced key is fetched once and lands on disk with the served bytes.
- rotate_proving_keys.sh uploads custom_ring.key with the rest of the set, and generate_lockfile.py includes sourced keys in the version hash, so a change to that key moves the immutable prefix instead of overwriting in place.
- The queue worker logs the real error of a failed custom ring job at error level before returning the scrubbed message, and the synchronous handler does the same. The only value-bearing error is the final public hash mismatch, both hashes are public.

Validation: Downloaded the published `custom_ring.key` from the current prefix and verified that its size and SHA-256 match the committed lockfile.

### DIFF
--- a/prover/server/prover/common/key_downloader.go
+++ b/prover/server/prover/common/key_downloader.go
@@ -300,14 +300,6 @@ func EnsureProvingKey(keyPath string, autoDownload bool, config *DownloadConfig)
 		}
 	}
 
-	if entry.Source != "" {
-		return fmt.Errorf(
-			"key %s is built from the %s artifact and is not on the object store: run just ensure-custom-ring-prover-key",
-			filename,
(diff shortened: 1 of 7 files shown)


sergeytimoshin:
Superseded by #312, which ports this behavior to current main and both current custom-ring circuits.

PR #296, opened 2026-09-04
fix: address circuit audit feedback 

ananas-block:
### Public transaction hashes expose spent UTXOs and dummy output positions

`private_tx_hash` now includes a private blinding derived from the transaction’s first nullifier and a private `TxSecret`. This prevents observers from enumerating candidate input commitments or dummy-output masks against the published digest.

Transact circuits derive the output blinding seed and private transaction blinding with separate domain tags. Merge circuits derive their private transaction blinding from the owner’s nullifier secret.

Addresses OSEC 2 and zkSecurity #01–#02.

### Duplicate transaction outputs enable Faerie Gold griefing

Every transaction output blinding, including padding, is constrained to a Poseidon derivation over the first nullifier, the output blinding seed, and the final physical output index. Reusing a seed across transactions or repeating output contents cannot reproduce the same commitment through blinding reuse.

Addresses the transact portion of zkSecurity #03. The deposit fix is separate.

### Proofs do not authenticate their input and output trees

Transact and both merge circuits now bind the input tree slots and output tree ID into the public-input hash. Each input selects its tree ID, UTXO root, and nullifier root together.

UTXO commitments also include the tree ID, separating otherwise identical UTXOs, nullifiers, and addresses across trees.

Implements circuit-side binding for OSEC 3 and zkSecurity #06; program and SDK integration remains outstanding.

### Batch-address templates can commit elements they never insert

`BatchAddressTreeAppendCircuit` validates its layout at the start of `Define`. Compilation rejects zero batch sizes or tree heights, mismatched element counts, and Merkle paths with incorrect lengths.

The validator adds no constraints to valid templates. Together with existing transact and merge path validation, it covers the production callers implicated by the Merkle-height finding.

Addresses OSEC 4 and 8.

### Address inputs bypass the nullifier-capacity policy

When `allow_dummy_inputs` is false, every transaction input must now be a real UTXO spend. Both dummy and address slots are rejected because they consume nullifier capacity without spending an existing UTXO.

Addresses OSEC 5.

### Mixed P-256 transactions expose the private owner identity

A ring P-256 spend cannot be combined with a default-ring P-256 spend or publish the shared P-256 identity through an output owner tag.

Address creation does not force owner publication: a ring P-256 spender can create an address privately, and an address-only proof publishes a zero default-owner hash.

Addresses OSEC 6.

### Dummy output tags can disclose private identities

A nonzero dummy output tag must repeat an identity already published by the transaction. Private output owner hashes no longer supply eligible dummy tags.

The payer is excluded from eligible signer identities, preventing padding from presenting a fee sponsor as a recipient.

Addresses OSEC 7.

### Signed-amount range checks accept an invalid boundary value

`RangeCheckSigned64` now rejects `-2^64`, restricting accepted amounts to `[-(2^64−1), 2^64−1]`, where the magnitude fits a `u64`.

Addresses OSEC 9.

### Limb comparisons assume equal operand widths

`IsLessLimbs` sizes each comparison using the wider corresponding operand. This preserves the bounded-comparison precondition when operands have different declared widths.

Existing callers use equal-width canonical limbs, so their constraint systems are unchanged.

Addresses OSEC 10.

### Owner identities lack signature-algorithm separation

The P-256 circuit now hashes the x-coordinate with the `0x50` algorithm prefix. It rejects the previous untagged identity and the Solana-tagged form; the signature message digest remains unchanged.

Implements the circuit portion of zkSecurity #08. The corresponding program and SDK identity derivations remain outstanding.

## Other

Transact and merge circuits support five input tree slots with private per-input selection. Selected slots must have nonzero roots. The slot commitment uses a right hash chain with a precomputable unused suffix.

The `ring-utils` proof now requires every input and output UTXO to belong either to the public ring program or to the default ring.

Transfer preparation derives blindings using final output positions, including compact layouts and padding. Low-level prover builders reject mismatched output blindings before proving.

Plaintext and anonymous sender decoding use the transaction’s first nullifier when reconstructing outputs. Missing nullifiers and output counts exceeding the widest supported shape return explicit errors.

Split bundles carry eight final output blindings instead of the derivation seed. The compression example persists its current blinding and supplies the spent blinding during updates.

Regression coverage exercises blinding derivation, privacy restrictions, malformed templates, arithmetic boundaries, tree selection, and public-input binding.

The local upgrade to gnark `v0.16.3` and gnark-crypto `v0.21.0` is **uncommitted**, so the GLV finding is not yet a committed branch fix. Final proving and verifying key regeneration is also outstanding for the latest circuit changes.


prover/server/circuits/spp_transaction/default/eddsa_only.go
```diff
@@ -20,10 +20,14 @@ type DefaultRingEddsaOnlyPublic struct {
...
 	OutputHashes []frontend.Variable
-	// UTXO tree roots to prove inclusion of real input UTXOs.
+	// Input tree slots: raw u16 tree ids and the UTXO tree roots to prove
+	// inclusion of real input UTXOs.
+	TreeIDs       []frontend.Variable
 	UtxoTreeRoots []frontend.Variable
-	// Nullifier tree roots to prove non-inclusion of input nullifiers.
-	NullifierTreeRoots []frontend.Variable
+	// Nullifier tree root to prove non-inclusion of input nullifiers.
+	NullifierTreeRoot frontend.Variable
```
sergeytimoshin: why only one NullifierTreeRoot?

prover/server/circuits/spp_transaction/custom/authority.go
```diff
@@ -17,17 +17,19 @@ import (
...
-	PublicAssets       [shared.NPublicSlots]frontend.Variable
-	PublicAmounts      [shared.NPublicSlots]frontend.Variable
-	RingProgramID      frontend.Variable
-	SignerPkHashes     []frontend.Variable
-	AllowDummyInputs   frontend.Variable
+	Nullifiers        []frontend.Variable
+	OutputHashes      []frontend.Variable
+	TreeIDs           []frontend.Variable
+	UtxoTreeRoots     []frontend.Variable
+	NullifierTreeRoot frontend.Variable
```
sergeytimoshin: why only one NullifierTreeRoot?

ananas-block:
superseeded by https://github.com/helius-labs/zolana/pull/298

PR #297, opened 2026-09-05
perf: enable 36 input UTXOs

ananas-block:
### Large consolidations exceed the supported circuit shapes

SPP now supports 36-in/2-out transfers on the confidential EdDSA, ring EdDSA, and ring P256 paths, plus 36-in/1-out merges on both merge paths. The prover, key generation, and on-chain verifying keys follow.

Large transfers require an explicitly declared shape. Merge builders pad to the smallest supported count, 8 or 36, while automatic wallet merges still select at most 8 inputs.

The large shapes require transaction v1. Existing SDK submission paths still build legacy messages, so submitting them requires a caller that builds v1 transactions.

### Parsing and proof buffers grow with the input count

Deposit, transact, and merge parsers now borrow validated lists directly from instruction data and decode entries as they are consumed. Account parsing borrows contiguous account slices, and proof assembly folds roots and output-owner hashes incrementally instead of retaining and copying per-entry arrays.

This reduces allocation and stack pressure as shapes widen. Signer identities remain deduplicated in first-occurrence order with the payer first, and proof assembly checks that every required component was assigned.

### External-data hashing rebuilds fields already present in the instruction

`transact` instruction data now places the external-data fields in one contiguous prefix, from `expiry_unix_ts` through `messages`. SPP hashes the instruction discriminator, that exact byte slice, and the committed settlement and output-owner account addresses in protocol order.

The commitment now covers the owner-tag encoding as well as any referenced address, so inline and account-backed representations of the same owner produce different hashes. Rust, TypeScript, Go, and the shared vectors follow. This changes both the instruction wire format and `external_data_hash`.

The TS external-data builder also keeps a private deep copy of committed fields and returns copies from byte-bearing getters. Mutating caller-owned arrays or the returned view can no longer change subsequent hashing or instruction assembly.

### Events duplicate transaction data as shapes widen

Transact events now contain only the first input queue sequence and first output leaf index, a fixed 16-byte body. Merge events use a fixed 48-byte body, retaining the output view tag needed to reconstruct registry-based merges.

Photon and the program-test indexer reconstruct nullifiers, outputs, messages, trees, and settlement information from the emitting instruction and its accounts. Reconstruction validates the parent instruction, event kind, shape, account layout, and index arithmetic. Photon’s rings parser version advances to 4.

### Prefunded nullifier PDAs exceed the instruction-trace limit

Nullifier PDA creation now uses `CreateAccountAllowPrefund`, requiring one System Program CPI for both empty and prefunded addresses. The previous allocate-and-assign path required two CPIs per prefunded input and could push a 36-input transaction beyond the 64-entry instruction-trace limit.

Forester fees are collected before direct tree-to-PDA lamport transfers, preventing an intermediate CPI from observing an unbalanced tree debit.

## Other

Tree creation validates initialization parameters before the first allocation or persistent mutation. Nullifier batch counters and update-event arithmetic use checked operations.

The Rust and TS default transact compute limit rises from 300,000 to 800,000 units to cover the larger shapes and P256 verification.

`xtask max-shape` and `max-merge-shape` measure serialized transaction sizes. Transaction-size tests, validator coverage for the 36-input transfer, event reconstruction tests, cross-language encoding vectors, and CU benchmarks follow.


program-libs/hasher/src/hash_chain.rs
ananas-block, line 1: revert changes in this file

program-libs/account-checks/src/account_iterator.rs
```diff
@@ -315,3 +346,31 @@ impl<'info> AccountIterator<'info> {
         let _ = (error, pubkey, expected, account_name, location);
     }
 }
+
+#[cfg(all(test, feature = "test-only"))]
+mod tests {
```
ananas-block, line 351: move to tests dir

program-libs/event/src/lib.rs
```diff
@@ -59,30 +50,46 @@ pub struct SplTransfer {
...
-    pub zkp_batch_size: u16,
-    pub old_next_index: u64,
-    pub start_sequence_number: u64,
-    pub first_root_index: u32,
-    pub num_update: u32,
-    pub first_zkp_batch_index: u32,
-    pub new_root: [u8; 32],
+#[derive(Clone, Copy, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize)]
+pub struct TransactEvent {
+    pub first_input_queue_seq: u64,
```
ananas-block, line 65: make this a vec to be compatible with multiple input trees

program-libs/event/src/lib.rs
```diff
@@ -59,30 +50,46 @@ pub struct SplTransfer {
...
+    pub first_output_leaf_index: u64,
+}
+
+impl TransactEvent {
+    /// Borsh encodes fixed-size fields as their little-endian concatenation, so
+    /// the whole body is this many bytes and can be written into a stack array.
+    pub const LEN: usize = 16;
```
ananas-block, line 72: why is this necessary?

program-libs/event/src/lib.rs
```diff
@@ -59,30 +50,46 @@ pub struct SplTransfer {
...
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize)]
+pub struct MergeEvent {
+    pub first_input_queue_seq: u64,
+    pub first_output_leaf_index: u64,
+    pub output_view_tag: [u8; 32],
+}
+
+impl MergeEvent {
+    pub const LEN: usize = 48;
```
ananas-block, line 92: why is this necessary?

PR #310, opened 2026-09-12
Fix sponsored ring withdrawals and bound RPC requests

sergeytimoshin:
Fix sponsored full ring withdrawals and prevent Solana RPC requests from waiting indefinitely.

- **Withdrawals:** When another account pays the fee and the entire selected ring balance is withdrawn, every output can be a dummy slot. The Rust and TypeScript SDKs currently reject this case because they expect a real output to determine the encoding length. Preserve the existing dummy-slot length so the withdrawal can be built.
- **RPC requests:** Add a configurable 30-second default timeout and support for a custom Solana Kit transport. Cancellation and timeouts stop waiting even when the transport ignores its abort signal; the wrapper does not retry requests.

Validation: SDK and Rust tests, strict packed-consumer checks, and CI pass, including TypeScript E2E and localnet suites.


custom-rings/sdk/src/transfer.rs
```diff
@@ -903,11 +903,12 @@ fn frame_dummy_outputs(proof_inputs: &mut SppProofInputs) -> Result<(), Transfer
             continue;
         }
         let in_ring = output.ring_program_id.is_some();
```
ananas-block, line 905: `in_ring` reads the dummy's own `ring_program_id`, and `finalize` builds dummies from `..Default::default()`, so it is `None` here. With a template present that does not matter, because the length comes from the real ring slot. With the new fallback there is no template, so the slot keeps the non-ring length and gets the non-ring scheme byte.

Measured with a temporary probe on this commit:

```
mixed ring transfer : real slot  len=121 scheme=4 (RingConfidential)
                      dummies    len=121 scheme=4
full withdrawal     : only slot  len=89  scheme=3 (Confidential)
```

The 32-byte gap is `ring_program_id: Option<Address>` in `ConfidentialOutputPlaintext`, which `dummy_ciphertext_len` (`sdk-libs/transaction/src/instructions/transact/transfer.rs:831`) encodes as `None`.

A ring transact that spends its whole balance now writes a slot 32 bytes shorter, under a different scheme byte, than one that leaves change. That labels the spend as "emptied the ring" for anyone reading Solana. Before this commit the transaction could not be built at all, so the distinguisher is new.

The tests pin the shape rather than catch it. `ring-full-withdrawal-v1.json` fixes `encodedLength: 89, scheme: 3`, and the assertion at line 1522 uses `confidential_encrypted_output_body` where `dummy_framing_matches_ring_slot_lengths` uses `ring_confidential_encrypted_output_body`.

Suggestion: take `in_ring` from the transfer's ring program id, which the builder already has, and size the fallback to an in-ring slot. Same change in `sdk-libs/ts/src/ring/transfer.ts:648-662`. If the current shape is a deliberate trade, say so at line 882; as written the doc comment reads as though nothing is lost.

sdk-libs/ts/src/client/kit.ts
```diff
@@ -77,6 +106,13 @@ function operationError(method: string, signal: ComposedSignal, cause: unknown):
       cause,
     });
   }
+  if (isClientError(cause)) return cause;
```
ananas-block, line 109: Moving `isClientError` below the timeout and abort branches is what makes "honors a shorter context deadline" pass, so the reorder is right.

It does break a contract `ring/program.ts` depends on. `refusedInPreflight` (`program.ts:705`) wraps a preflight refusal in a `ClientError` with `reason: "transaction failed"` so `retryable()` returns false, and its doc comment says `runKitRpc` leaves it alone. If the composed signal has timed out in the same tick, that error now returns as `CLIENT_TIMEOUT`, `retryable()` (`program.ts:719`) returns true, and the deploy loop resubmits a transaction that already failed preflight.

Either test `isClientError(cause) && cause.details?.["reason"] === "transaction failed"` ahead of the timeout branch, or update that comment.

sdk-libs/ts/src/client/kit.ts
```diff
@@ -29,15 +33,47 @@ export function createKitClients(
...
     solanaRpcUrl: string | URL;
     solanaRpcSubscriptionsUrl?: string | URL;
+    solanaRpcTransport?: RpcTransport;
+    solanaRpcRequestTimeoutMs?: number;
   }>,
 ): Readonly<{ solanaRpc: SolanaRpc; solanaRpcSubscriptions: SolanaRpcSubscriptions }> {
   const rpcUrl = urlString(input.solanaRpcUrl, "solanaRpcUrl", ["http:", "https:"]);
   const subscriptionsUrl =
     input.solanaRpcSubscriptionsUrl === undefined
       ? defaultSolanaRpcSubscriptionsUrl(rpcUrl)
```
ananas-block, line 43: `solanaRpcUrl` still feeds `defaultSolanaRpcSubscriptionsUrl`, and `resolveClientEndpoints` (`src/endpoint.ts:42`) defaults it to the localnet endpoint. So `new ZolanaClient({ solanaRpcTransport })` sends RPC through the transport while subscriptions open against `ws://127.0.0.1:8900`, with the indexer and prover URLs defaulting to localnet as well.

`scripts/dist-consumer.ts:49` shows the trap: it hardcodes `solanaRpcUrl: "https://rpc.example"` beside a real transport, which points subscriptions at a host that does not resolve.

Worth requiring `solanaRpcSubscriptionsUrl` when a transport is supplied, or covering it in the new README section, which does not mention subscriptions.

custom-rings/sdk/src/transfer.rs
```diff
@@ -1440,6 +1441,97 @@ mod tests {
...
+        let slots =
+            encode_confidential_slots(&prepared.outputs, &AssetRegistry::default(), &key, salt)
+                .expect("slots");
+        prepared
+            .finalize(key.pubkey(), salt, slots)
+            .expect("finalize")
+    }
+
+    #[test]
+    fn dummy_framing_sponsored_full_withdrawal_matches_typescript_vector() {
```
ananas-block, line 1488: I set the fee payer to the sender's own address and reran: outputs are still all dummy and the same path runs. `is_dummy()` is `owner_address.is_none()`, which the payer does not affect.

The test name, the CHANGELOG entry, and the commit subject all scope this to a relayer-paid withdrawal. Someone hitting `InvalidDummyOutput` on a self-paid full withdrawal will not find it. Drop the word from the three.

Separately, `matches_typescript_vector` overstates what runs. The vector is three numbers shared by both languages, and no TypeScript behaviour is checked. `matches_shared_vector` is closer.

One more on this block: `include_str!("../../../sdk-libs/ts/vectors/...")` at line 1489 reaches from a Rust crate into the TypeScript package directory. Harmless while `publish = false`, but it breaks quietly if that directory moves. And the parity assertions at line 1495 onward check `payer` plus an indirect `restored.hash()` comparison, where the TypeScript twin checks `inputUtxos`, `blindingSeed`, `outputTreeId`, `outputs`, and nine `externalData` fields. For a mirrored pair the Rust side should cover the same list.

custom-rings/sdk/src/transfer.rs
```diff
@@ -1440,6 +1441,97 @@ mod tests {
...
+            .zip(&original.external_data.outputs)
+        {
+            assert_eq!(encoded.utxo_hash, before.utxo_hash);
+            assert_eq!(encoded.owner_tag, before.owner_tag);
+            let data = encoded.data.as_ref().expect("framed dummy");
+            assert_eq!(
+                data.len() as u64,
+                vector["encodedLength"].as_u64().expect("length")
+            );
+            assert_eq!(data[5] as u64, vector["scheme"].as_u64().expect("scheme"));
```
ananas-block, line 1521:
```rust
assert_eq!(data[5] as u64, vector["scheme"].as_u64().expect("scheme"));
```

`data[5]` panics on a short slice, and the offset arrives with no derivation. `confidential_encrypted_output_body(data)` on the next line already returns the body whose first byte is the scheme, which is how the TypeScript twin reads it through `readOutputData`. Lines 1526 and 1531 index `outputs[0]` the same way; `.first()` / `.first_mut()` with `expect` keeps the repo rule against `[i]` in tests.

sdk-libs/ts/CHANGELOG.md
```diff
@@ -432,6 +435,11 @@ Changed
 
 Fixed
 
+- `buildRingWithdrawalTransaction` accepts a sponsored withdrawal of the entire
```
ananas-block, line 438: The entry credits `buildRingWithdrawalTransaction`, but both new tests call the framing helper directly. The one end-to-end withdrawal (`test/e2e/ring-flow.live.test.ts:295`) takes a fixed `5_000n` and leaves change, so no test proves or lands a full withdrawal.

An all-dummy output set is a shape no other test covers: the auditor decrypts nothing, and `waitForAudited` has no output to find. A builder-level case would be worth adding before the entry claims the builder accepts it.

sdk-libs/ts/src/services/signal.ts
```diff
@@ -40,3 +40,29 @@ export function composeSignal(context: RequestContext | undefined): ComposedSign
...
+    onAbort = () => reject(aborted());
+    signal.addEventListener("abort", onAbort, { once: true });
+  });
+  try {
+    const work = Promise.resolve().then(() => {
+      if (signal.aborted) throw aborted();
+      return operation();
+    });
+    const result = await Promise.race([work, cancellation]);
+    if (signal.aborted) throw aborted();
```
ananas-block, line 63: Minor, batched here:

- This post-race abort check drops a result that already arrived. Harmless for `sendTransaction`, where the signature comes from `getSignatureFromTransaction` and `landedSignature` recovers it. On a read it turns a `getLatestBlockhash` that resolved in the same tick as the 30s timer into `CLIENT_TIMEOUT`.
- Line 59 is unreachable. Between the entry check at line 51 and that microtask only synchronous caller code runs, and aborts arrive as macrotasks.
- `RequestContext.timeoutMs` at line 12 has no upper bound, so a value above 2^31-1 overflows `setTimeout` and fires at once. The new `solanaRpcRequestTimeoutMs` check caps exactly that; the same cap here would be cheap. Pre-existing.
- In `test/ring-transfer.test.ts`, `expect(damaged.externalData.outputs).toEqual(outputs)` cannot fail, because `frameDummyOutputs` returns a new `SppProofInputs` without mutating. The Rust equivalent does test something, since that function mutates in place.

ananas-block (commented):
Reviewed `54b2299dd`.

The fallback is the right shape. `finalize` writes 89 bytes of raw random for a dummy, which is not a parseable `OutputDataEncoding` frame (scheme bytes of 133 and 172 in a probe I ran before framing), so the framing pass does real work on this path. The `operationError` reorder is justified, and the test that forced it is named for the behaviour it pins. The RPC tests are adversarial where it counts: a transport that ignores its signal, and a late success after cancellation.

Two things to resolve before merge: the full-withdrawal slot is distinguishable from a ring slot (comment on `custom-rings/sdk/src/transfer.rs:905`), and the reorder changes retry behaviour after a preflight refusal (comment on `sdk-libs/ts/src/client/kit.ts:109`).

Ran `cargo test -p custom-ring-sdk --lib dummy_framing` (2 passed), `vitest run test/rpc-transport.test.ts test/ring-transfer.test.ts` (47 passed), `cargo clippy -p custom-ring-sdk --all-targets` (clean), `npm run lint` and `changelog:check` (clean). `npm run typecheck` and `format:check` fail on files this commit does not touch. My local `@solana/kit` is 7.0.0 against the declared `^8.3.0`, so the TypeScript run did not exercise 8.3.

PR #312, opened 2026-09-13
Serve custom ring keys from object store

sergeytimoshin:
The prover now downloads both custom-ring proving keys from the same versioned object-store prefix as other proving keys. Custom-ring failures keep a scrubbed client response while logging the underlying server error.

Validation: `go test ./...` in `prover/server`; both published objects match the lockfile size and SHA-256.


prover/server/scripts/generate_lockfile.py
```diff
@@ -54,7 +54,7 @@ def main() -> int:
         action="append",
         default=[],
         metavar="NAME",
-        help="key built from release assets, pinned but never served from the object store",
+        help="key built from release assets, recorded as its provenance",
```
ananas-block, line 57: **`just regen-custom-ring-keys` now produces a key set nobody can fetch**

`generate_keys_custom_ring.sh:33` regenerates both custom-ring keys — gnark's Setup is non-deterministic, so the bytes change on every run — and then calls this script with `--only-release`. `pin_release_entries` rewrites the two sha256/size entries and never touches `manifest["prefix"]`. I ran it against a copy of the committed lockfile to confirm:

```
prefix after --only-release: proving-keys/7765d8fa45c5416c   <- unchanged
custom_ring_base entry     : sha256 e59d9182...              <- new bytes
```

That was coherent before this PR: the two keys were not on the object store, the justfile pulled them from the hand-bumped `custom-ring-keys-v7` tag, and the prefix hash excluded them by construction. Now the downloader resolves them to `<base>/<prefix>/<name>`, a folder the design calls immutable, which still holds the old bytes. Anyone without a local copy gets a checksum hard-fail out of `downloadAndVerify`, and `just ensure-custom-ring-live-keys` dies with `hashes to X, proving-keys.lock pins Y`. The only way out is overwriting objects inside an immutable version folder.

There is no safe use of `--only-release` left after this change. Suggest deleting it along with `pin_release_entries`, and pointing `regen-custom-ring-keys` at a full regeneration so the prefix moves with the keys.

`rotate_proving_keys.sh`, `rotate_local_no_upload.sh` and `just build-spp-keys` all call the full generator, so they recompute the prefix and are unaffected.

prover/server/scripts/generate_lockfile.py
```diff
@@ -104,7 +104,7 @@ def main() -> int:
     # versions untouched. 16 hex chars (64 bits) is collision-safe across the
     # handful of key-set versions this project will ever have.
     canonical = json.dumps(
-        {name: entry["sha256"] for name, entry in keys.items() if "source" not in entry},
+        {name: entry["sha256"] for name, entry in keys.items()},
```
ananas-block, line 107: **The committed lockfile's prefix no longer matches its own generator**

The canonical set now includes `source`-marked keys, but `proving-keys.lock` is net-unchanged in this PR. Recomputing over the committed key set:

```
committed prefix          : proving-keys/7765d8fa45c5416c
old rule (exclude source) : 7765d8fa45c5416c
new rule (include source) : f6a170e063f0a347
```

The module docstring ("the version hash is derived from the whole key set") and the comment three lines above this one ("it changes iff any key changes") are both false for the artifact that ships.

Nothing breaks today, because `prefix` is read verbatim on both the upload and the download side. The cost is deferred: the next full rotation jumps the prefix for a reason unrelated to any key changing, and anyone auditing the lockfile by regenerating it gets a mismatch with no way to tell whether a key moved.

There are two consistent options. Either regenerate the lockfile in this PR and upload the full key set to `f6a170e063f0a347`, or leave the canonical set excluding `source` and revert this line. Please do not leave the generator and the committed artifact disagreeing.

ananas-block (commented):
The fetch path works. `custom_ring_base.key` and `custom_ring_policy.key` both return 200 under `proving-keys/7765d8fa45c5416c/` with `content-length` matching the pinned sizes (58012883 / 145691406). `go vet` and `go test ./prover/common/...` are clean.

The regeneration path is the problem. Two changes together break the invariant that the version hash is a function of the key set:

- `--only-release` repins the custom-ring hashes while keeping the prefix, which was safe only while those keys lived outside the object store.
- the canonical hash now covers `source`-marked keys, but the committed lockfile was not regenerated, so its prefix is the old-rule value.

Blocking on those two. The rest is cleanup that falls out of deleting the now-dead `Source` field.

prover/server/prover/provingkeys/provingkeys.go
```diff
@@ -21,7 +21,7 @@ var embeddedLockfile []byte
 type Entry struct {
 	Sha256 string `json:"sha256"`
 	Size   int64  `json:"size"`
-	// Names where a key comes from when the object store does not carry it.
+	// Provenance of a key whose setup also ships as a release asset.
```
ananas-block, line 24: **`Source` is dead now — delete it rather than re-document it**

With the gate gone from `EnsureProvingKey`, nothing in production Go reads this field. Grep across `provingkeys/` and `common/` finds three hits: this declaration, and two lines in `key_downloader_test.go`. The Python script writes it, the lockfile stores it, the struct parses it, and no code reads it.

Softening the comment to "Provenance of a key whose setup also ships as a release asset" keeps the field alive as documentation, which is the pattern we normally delete. Suggest removing the field, `--release` / `--only-release`, `pin_release_entries`, and the `--release` arguments at the three call sites (`rotate_proving_keys.sh:81`, `rotate_local_no_upload.sh:65`, `generate_keys_custom_ring.sh:33`) in one go.

If provenance is worth recording for humans, put it in the rotation docs. The lockfile is generated.

prover/server/prover/common/key_downloader_test.go
```diff
@@ -279,25 +279,31 @@ func TestEnsureProvingKeyNotInManifestMissingErrors(t *testing.T) {
 	}
 }
 
-func TestEnsureProvingKeySourcedKeyIsNotFetched(t *testing.T) {
+func TestEnsureProvingKeyWithSourceIsFetched(t *testing.T) {
```
ananas-block, line 282: **This test does not exercise `Source`**

It is `TestEnsureProvingKeyDownloadsMissingKey` with a different filename plus one inert `entry.Source = "release"`. Since no production code reads `Source`, deleting that line changes nothing about what the test proves.

It is a legitimate regression guard for the removal (it fails if `key_downloader.go` is reverted), but only because the dead field still exists. If `Source` goes, this collapses into its sibling and should be deleted with it. If the field stays for some reason I am missing, the name should say what it pins.

prover/server/server/queue_job.go
```diff
@@ -724,6 +724,14 @@ func (w *BaseQueueWorker) processMergeProof(payload json.RawMessage, circuitType
 }
 
 func (w *BaseQueueWorker) processCustomRingProof(payload json.RawMessage, circuitType common.CircuitType) (*common.Proof, error) {
+	proof, err := w.customRingProof(payload, circuitType)
```
ananas-block, line 727: **Good catch on the lost error, but this patches one of four holes**

The underlying bug is at line 479: `err = w.redactProofError(err)` runs before the log, so on `zk_custom_ring_queue` the real cause never reaches the logs. Wrapping `processCustomRingProof` recovers one source of that error. Still silently redacted for the same queue:

- line 410, the panic message. `w.redactProofError(fmt.Errorf("panic: %v", r))` redacts before logging, so a custom-ring panic logs as "custom ring proof failed" with the stack detail gone, which is where the missing detail costs most.
- `generateProof`'s three pre-switch errors: meta parse failure, `circuit %s cannot run on %s`, and unknown circuit type.

Logging the raw `err` at 479 before redacting, and the raw panic at 410, covers all four in one place and makes this wrapper method unnecessary. Redaction stays where it is for anything client-visible.

PR #316, opened 2026-09-14
feat: additional controls for custom rings

Atamanov:
This branch adds a set of ring-level controls on top of the audited transfer: a co-signer, a permanent delegate, and transfer limits with flow control.
Everything is enforced on-chain without new trust. Public amounts are read straight from the transaction's settlement legs. Private amounts are compared inside the policy circuit and handed to the program as single yes/no bits, so the program never sees the hidden value. Per-sender velocity is tracked in a compressed, member-keyed spend record, and the indexer reconstructs and serves those records. The delegate runs on SPP's authority rail, so it can move ring value exempt from member limits while every move stays audited.

- Co-signer (dual control): the ring authority names a second Solana signer and the operation classes it must approve — transfers, deposits, withdrawals . All with per-mint amount thresholds on withdrawals. Above a private-amount threshold the proof itself demands the signer.
- Permanent delegate: the program upgrade authority appoints one standing delegate, set once and never rotated. It moves ring value on the authority rail, exempt from velocity caps and counters, internal-only, and every move is audited.
- Per-transfer caps: each transfer's outflow is bounded per asset and refused before it is sent.
- Windowed velocity caps: each sender's cumulative outflow is bounded over a fixed window, tracked in a compressed spend record consumed and replaced atomically with the transfer.
- Public spend windows: per-mint deposit and withdrawal caps over a fixed window, enforced on the public legs.
- Compressed spend records and head map: member velocity state lives in one member-keyed Merkle tree with only its root on-chain, and Photon rebuilds the tree and serves membership witnesses.
- Auditor spend records: the auditor sees each transfer's velocity counters alongside the transfer.
- Flow control between features: scope and amount are independent reasons to demand the co-signer. A signature never makes an over-cap member transfer valid. The delegate is a separate rail that bypasses velocity, not an override of the cap.
- New proof circuits: a compressed policy circuit (member transfer with the head-map transition), a compressed registration circuit, and a delegate policy circuit, each with committed verifying keys and fingerprint guards.
- Client and operator support: Rust and TS SDKs and the CLI expose configuration, instruction builders, and the delegate and co-signer flows.


program-libs/event/src/ring_deposit_audit.rs
ananas-block, line 1: please move into another crate outside of program libs

sdk-libs/transaction/src/instructions/ring_authority.rs
ananas-block, line 1: this should be in a ring sdk imo

ananas-block:
The reason given in 67655cc0f ("A ring deposit could forge a namespace-owned zero-counter record") doesn't justify the head map. That forgery is possible because the shielded pool's deposit path doesn't require the owner to sign a data-carrying output; transact already does (`ConstrainOutput`). The fix belongs in the shielded pool and is coming in a separate PR. Once it lands, a member can't forge a record, so we can drop the head map, its root account and the Photon projection. That also ends the ring-wide root conflicts between transfers.

ananas-block:
4623d64c7 limits policy proofs to the current nullifier root (`NULLIFIER_ROOT_WINDOW = 0`). That doesn't make revocation safe: a queued nullifier isn't in the current root either. It also makes every transfer race the forester. The fix should follow the pool: accept any root in the history, and check that the looked-up entries' nullifier PDAs don't exist. The nullifier tree only lets a PDA close once no accepted root can still prove its absence, so the two checks together are complete.

ananas-block:
Nothing checks the escrowed nullifier key. The register-key circuit proves that some secret is encrypted to the auditor, but nothing ties that key to the member's notes. No transfer or deposit checks that notes are sent to a registered key (enrolling on first use happens only in the SDK). A member can register a random key, or none, and the delegate can never spend notes sent to them. When a delegate is set, every transfer and deposit should prove that each output's nullifier public key is in the key registry.

Open question: what happens with recipients who haven't registered? One option is to allow nullifier key = 0 for them, the same zero secret PDA owners already use. The delegate could then always compute their nullifiers, at the cost of their notes' spent status being public.

ananas-block:
**Problem:** the ring pins one entries tree for its whole life (`create_policy`). Windowed velocity rings also force every input, output, deposit, merge and delegate move into it (`transact.rs:489-490`, `forward.rs:107`). When that tree is full, lists can no longer change, and windowed members can no longer transfer or withdraw.

**Cause:** the policy circuit proves everything against a single tree's root pair.

**Only address creation needs a fixed tree.** Uniqueness has to be proven against one nullifier tree, so:
- creating an address (a new entry or spend record) uses the pinned address tree. Creation can involve two trees: the fixed address tree for the address claim, and any tree to store the new state in;
- an absence proof for an entry that was never created also checks that tree.

**Reading existing state only needs the tree that holds it:**
- inclusion under that tree's state root;
- non-inclusion of the nullifier under that tree's nullifier root;
- the nullifier PDA doesn't exist.

**Fix:** let entries, spend records and money live in, and move to, any tree, as the compression example's `update` does. The policy circuit takes one root pair per tree slot, like the pool's inputs, and pins only the address tree.

alex-helius:
> The reason given in 67655cc0f ("A ring deposit could forge a namespace-owned zero-counter record") doesn't justify the head map.

Fixed in e9e66d50a by removing the head map, its root account and its Photon projection, with the current record served by `getRingSpendRecord` and forgery closed by the deposit owner check in #315.

alex-helius:
> 4623d64c7 limits policy proofs to the current nullifier root (`NULLIFIER_ROOT_WINDOW = 0`).

Fixed in 88c365b48 by accepting any live nullifier root in the history, with the existing revocation target nullifier PDA check covering queued and inserted revocations.

alex-helius:
> Nothing checks the escrowed nullifier key.

Fixed in b30620001 and 6ec6bf59a. `set_delegate` now requires a policy ring and a key registry and turns key escrow on for good. From then on the policy, delegate and deposit circuits require every UTXO output's nullifier key to be registered for its owner under a `KeyRegistryRoot` history root. On the open question, the zero key passes only for the ring's namespace-owned spend records. A zero-key note sent to anyone else hides its owner hash, and the delegate cannot move it. Unregistered recipients cannot receive on an escrowed ring, and plain deposits are refused there.

alex-helius:
> **Problem:** the ring pins one entries tree for its whole life (`create_policy`).

Fixed in 6ec6bf59a and b30620001. The policy circuit takes one root pair per tree slot and pins only `address_tree`. Address claims and never-created absence proofs use that tree, and every other read uses the tree that holds the state. Entries, spend records and money live in any tree and move between trees on writes. Each revocation nullifier PDA derives under its fact's tree.

ananas-block:
The ring projection adds 5 Photon migrations, and 4 of them are this branch's own history: `head_maps` is created and later dropped, and `unify_ring_projections` only moves data between earlier layouts. None has run against a deployed database, so please squash them into one migration that creates the final tables:

- `ring_projection_{blocks,cursor,pending}`
- the key-registry `roots` and `members` tables
- `ring_spend_records` (with its index) and `ring_spend_record_rings`

Then drop the other four and the in-between upgrade tests in `migration/mod.rs`. Keep it in `RingsMigrator` and unconditional rather than gated on `ring-projection`, like the other ring tables, so a database never hits "applied migration missing" when builds switch features.

ananas-block:
The TS package exports the key registry's tree helpers under names from the removed head map:
- `HEAD_MAP_HEIGHT`, `HEAD_MAP_CAPACITY`
- `headMapLeaf`, `headMapRootFromProof`, `verifyHeadMapInsert`
- the `HeadMapLeaf`, `HeadMapPath` and `HeadMapInsertProofInput` types
- `CustomRingHeadInsertion`
- the error code `RING_HEAD_MAP_INVALID`

None of these are on `main` or in any `CHANGELOG.md` entry, so they most likely never shipped. Please rename them after the key registry before 0.3.0-alpha ships, as `991e8e15f` did on the Go side. Then list them in the unreleased entry: `AGENTS.md` requires every public change to update `CHANGELOG.md` under `CHANGELOG-RULES.md`.

ananas-block:
`docs/spec.md:217` still says "All sources live in the transfer's entries tree." Since `b30620001`, entries can live in any policy tree and only their addresses are claimed in the pinned address tree. Suggestion: update the spec to match.

ananas-block:
The key registry still uses head-map names in Rust and in the circuit's public fields. Suggested renames, following `KeyRegistryRoot`, `RegisterKeyIxData.registry_old_root` and `6ec6bf59a`'s `registry.Leaf` / `registry.Insertion`:
- `custom_ring_interface::head_map` → the `key_registry` module: `KEY_REGISTRY_HEIGHT` / `_CAPACITY` / `_EMPTY_ROOT`, `KeyRegistryLeaf { member, next, key_hash }` (the value is `Poseidon(nullifier_pk, ct_hash)`, not a nullifier), `KeyRegistryInsert`, `KeyRegistryVerifyError`, `KeyRegistryTransition`, and `RegisteredKey::commitment()` → `hash()` (`NAMING.md` reserves "commitment" for BSB22)
- `zolana-ring-head-map` → `zolana-ring-key-registry`, with `HeadMap` → `KeyRegistryTree`
- Go: `KeyRegisterCircuit.HeadOldRoot` / `HeadNewRoot` → `RegistryOldRoot` / `RegistryNewRoot` and `registry.Leaf.Key` → `KeyHash`, with the witness keys, `head_insertion.go` and the Rust witness writers moved in one commit, as `NAMING.md` asks

Separately, `HeadTransfer`, `TransferProofInputs` and `HeadMap::transfer` are left over from the per-transfer head update. Their only user is the SDK test in `key_registry.rs`, which runs a no-op transfer to get an inclusion proof. Suggestion: replace them with a `member_proof(member)` method.

ananas-block:
The `// … retired.` markers for 8157, 8160–8163, 8165–8167 and 8171 cover codes that were added and removed on this branch and never reached `main`; only 8129 did. Suggestion: unless a devnet deployment exposed them, drop those markers and the `retired` guard in `error_codes_are_stable`.

ananas-block:
A few Photon leftovers besides the migrations:
- `services/photon/README.md:131` describes the spend-record migration dropping the head-map tables, which will be wrong once the migrations are squashed.
- `RingsTreeKind::KeyRegistry` says "Only the current root is accepted", but the registry root now keeps a 32-root history for escrow proofs.
- `ProjectionKind` is left with one variant, `KeyRegistry`.

Suggestion: update the README and the doc comment, and consider dropping `ProjectionKind`.

ananas-block:
With the head-map root gone, `AppendRoot` has one implementer, `KeyRegistryRoot`, but `process_create_indexed_root_ix`, `load_append_root_mut`, `SentinelRootInit::init` and `RootTransition::apply` are still generic over it. Suggestion: make them concrete and drop the trait.

ananas-block:
Nit: `head_roots` (`custom-rings/sdk/src/witness.rs:823`, mirrored in TS comments) returns the tree account's current roots. Suggestion: with the head map gone, `current_roots` would avoid reading it as head-map roots.

alex-helius:
> The ring projection adds 5 Photon migrations, and 4 of them are this branch's own history

Fixed in 545767455. One unconditional `RingsMigrator` migration, `m20260923_000001_ring_projection`, creates the seven final tables and the spend record index, and the other five migrations and the upgrade tests are gone. The same commit drops `ProjectionKind` and makes the `Projection` trait, `RingStore` and the indexer's `OnChainRoot`, `Leaf`, `Append` and `Spliced` concrete.

alex-helius:
> The TS package exports the key registry's tree helpers under names from the removed head map

Fixed in 62265b317. They are `KEY_REGISTRY_HEIGHT`, `KEY_REGISTRY_CAPACITY`, `keyRegistryLeaf`, `keyRegistryRootFromProof`, `verifyKeyRegistryInsert`, the `KeyRegistryLeaf`, `KeyRegistryPath` and `KeyRegistryInsertProofInput` types, `CustomRingRegistryInsertion` and `registeredKeyHash`, and `RING_HEAD_MAP_INVALID` folds into `RING_KEY_REGISTRY_INVALID`. The 0.3.0-alpha entry lists them, and 3e0822042 also drops an entry that fixed a never-shipped spend record bug and brings the entry under the twenty item ceiling.

alex-helius:
> `docs/spec.md:217` still says "All sources live in the transfer's entries tree."

Fixed in 09fe11d99. It now says every source claims its entry addresses in the ring's address tree, and the entries live in any tree. The same commit fixes the equivalent sentences in `custom-rings/README.md`, `docs/ring-policy*.md` and the wizard hint.

alex-helius:
> The key registry still uses head-map names in Rust and in the circuit's public fields.

Fixed in 62265b317:
- The `key_registry` module holds `KEY_REGISTRY_HEIGHT`, `KEY_REGISTRY_CAPACITY` and `KEY_REGISTRY_EMPTY_ROOT`, plus `KeyRegistryLeaf`, `KeyRegistryInsert`, `KeyRegistryVerifyError`, `KeyRegistryTransition` and `RegisteredKey::hash()`.
- The crate is `zolana-ring-key-registry`, with `KeyRegistryTree`.
- In Go, `KeyRegisterCircuit` has `RegistryOldRoot`, `RegistryNewRoot` and `LowKey`. The witness keys, `registry_insertion.go` and the Rust and TS writers moved in the same commit.
- The register_key proving key keeps its pk and vk, and only its stored constraint system changed. It ships as `custom-ring-keys-v14`, and the committed verifying key is unchanged.
- `HeadTransfer`, `TransferProofInputs` and `HeadMap::transfer` are gone. The SDK test uses `KeyRegistryTree::member_proof(member)`.

alex-helius:
> The `// … retired.` markers for 8157, 8160–8163, 8165–8167 and 8171 cover codes that were added and removed on this branch

Fixed in 9acfd7449. No released binary returned a code above 8141, so the branch codes are compacted to 8142..8172 with no markers and no guard. 8123, 8128 and 8129 keep main's markers. The same commit compacts the instruction tags (21..31), the account discriminators (7 and 8), the Photon ring RPC codes (-32070..-32074) and the key registry tree kind (3), because none of them shipped either.

alex-helius:
> With the head-map root gone, `AppendRoot` has one implementer, `KeyRegistryRoot`

Fixed in a2f9d2c7c. The generic code is replaced by `KeyRegistryRootInit`, `load_key_registry_root_mut` and `process_create_key_registry_root_ix`. `RootTransition::check` runs before proof verification, and the resulting `CheckedRootTransition::apply` runs after it, so the stale root and cursor checks run once. `KeyRegistryRoot` also drops its duplicate `root` field, and the current root is `history[history_cursor]`, read through `root()`.

alex-helius:
> Nit: `head_roots` (`custom-rings/sdk/src/witness.rs:823`, mirrored in TS comments)

Fixed in 62265b317. It is now `current_roots`, and the TS mirror comments follow. The shipped TS `TreeHeadRoots` and `decodeTreeHeadRoots` keep their names.

custom-rings/interface/src/deposit_audit.rs
alex-helius: Moved out of program-libs in 33aa04585, the capsule decoding lives in `custom-rings/interface` and no ring code is left in `program-libs/event`.

custom-rings/sdk/src/prepared_authority.rs
alex-helius: Moved into the ring SDK in d8faa45d4 and 753f9bae6, `sdk-libs` keeps only the generic `attach_input_proofs` and `RingAuthorityProver`.

ananas-block (approved):

PR #317, opened 2026-09-15
feat: expose TypeScript user registry instruction builders

tilo-14:
## Summary

Expose public user-registry PDA and instruction builders so registration and merge opt-in can share one transaction. Reuse the builders in existing wallet helpers while preserving registration behavior, key updates, and P-256 restrictions.

Rebased onto `e5e3fe8521dc0bf39ac77b60fae38f4aba5418ef`, the source of the latest published `@heliuslabs/zolana@0.1.6-alpha`. Target `release/dev-v2`, which contains that package source. The `0.1.7-alpha` version and changelog entry are unreleased; no package has been published.

## Validation

- Full SDK `npm run check`: formatting, lint, changelog, typecheck, 591 tests, build, and distribution checks passed.
- Rust shared-wire fixture test passed.
- Package dry run passed; the packed SDK passed consumer typechecking with `skipLibCheck: false` and unused-code checks.
- TypeScript registration/merge example passed on devnet: registration and merge opt-in in one transaction, consolidation, stale two-device merge rejection, resync, and rebuilt retry.
- Existing TypeScript deposit/transfer/withdraw example passed on devnet after migration to the current `LocalKeys` API.


sdk-libs/ts/src/interface/instructions/index.ts
```diff
@@ -50,6 +53,58 @@ type Meta = NonNullable<Instruction["accounts"]>[number];
...
+  readonly nullifierPublicKey: Bytes32;
+  readonly viewingPublicKey: Bytes33;
+}
+
+function registryKeysData(tag: number, data: RegisterInstructionData): Uint8Array {
+  if (data.ownerP256 !== undefined) {
+    fail("INTERFACE_CODEC", { reason: "P256 owner registration is unsupported" });
+  }
+  return new Writer()
+    .u8(tag, "discriminator")
```
ananas-block, line 67: The discriminators are bare numbers here and at lines 78, 92 and 102. The other builders in this file name their tags (`InstructionTag.deposit`, `InstructionTag.pauseTree`), Rust names these in `instruction::discriminator::{REGISTER, SET_MERGING_ENABLED, UPDATE_KEYS}`, and this PR deletes a named `const SET_MERGING_ENABLED = 1` from `registry.ts` to get here. Net result is one fewer named constant than before.

Could we add a frozen tag object next to `InstructionTag` in `interface/program.ts`?

```ts
export const UserRegistryInstructionTag = Object.freeze({
  register: 0,
  setMergingEnabled: 1,
  updateKeys: 2,
} as const);
```

sdk-libs/ts/src/interface/instructions/index.ts
```diff
@@ -50,6 +53,58 @@ type Meta = NonNullable<Instruction["accounts"]>[number];
...
+    .u8(0, "ownerP256")
```
ananas-block: `Writer.option()` (`interface/internal.ts:131`) already writes the Borsh option tag. Using it here would keep the `None` distinguishable from the discriminator `0` on the line above.

sdk-libs/ts/src/interface/instructions/index.ts
```diff
@@ -50,6 +53,58 @@ type Meta = NonNullable<Instruction["accounts"]>[number];
 
 export type SignerAccount = Address | TransactionSigner;
 
+export interface RegisterInstructionData {
+  readonly ownerP256?: undefined;
```
ananas-block: `ownerP256?: undefined` is a property that can never hold a value, so the `fail("INTERFACE_CODEC", ...)` guard below is unreachable from TypeScript. The test needs `@ts-expect-error` to reach it, and the README has to tell users to write `ownerP256: undefined` explicitly.

The runtime guard earns its keep: silently registering a record with no P-256 owner would be worse than throwing for a JS caller who passes one. Whether the field belongs in the TS type is the open part. Dropping it and keeping the guard as a check on untyped input gives the same protection with one less phantom in the public surface. Happy either way, just want it to be a decision rather than an artifact of mirroring `RegisterData`.

sdk-libs/ts/src/interface/instructions/index.ts
```diff
@@ -50,6 +53,58 @@ type Meta = NonNullable<Instruction["accounts"]>[number];
...
+  input: Readonly<{ userRecord: Address; owner: SignerAccount; enabled: boolean }>,
+): Instruction {
+  return instruction(
+    new Writer().u8(1, "discriminator").bool(input.enabled, "enabled").finish(),
+    [meta(input.userRecord, false, true), meta(input.owner, true, false)],
+    USER_REGISTRY_PROGRAM_ID,
+  );
+}
+
+export function updateRegistryKeysInstruction(
```
ananas-block, line 103: `updateRegistryKeysInstruction` is defined here but re-exported by neither `src/instructions.ts` nor `src/interface/index.ts`, so it stays internal to `registry.ts` while `register` and `setMergingEnabled` become public. For a PR titled "expose TypeScript user registry instruction builders", is the omission deliberate?

Key rotation is still reachable through `buildRegistrationTransaction`. It just cannot be included in a caller's own transaction the way the other two now can. If that is intended, the registration paragraph in the README is the place to say so.

sdk-libs/ts/src/wallet/registry.ts
ananas-block, line 1: `publishedKeysMatch` (line 366, outside the diff hunk) still computes and compares `ownerP256`, but it is reachable only from `registrationInstruction`, which runs after the `WALLET_P256_REGISTRATION_UNSUPPORTED` guard in `buildRegistrationTransaction`.

Removing the mirrored branch from `registrationInstruction` is right. The old code set the `ownerP256 = 1` flag without the Instructions sysvar or the preceding secp256r1 proof-of-possession, so `process_register` would have rejected it. This is the other half of the same dead pair; worth deleting in the same change.

sdk-libs/ts/test/user-registry.test.ts
```diff
@@ -0,0 +1,300 @@
...
+  it("returns undefined without a blockhash when the published keys already match", async () => {
+    const { shieldedAddress, record } = registrationFixture();
+    const client = { getAccount: vi.fn(async () => record), getLatestBlockhash: vi.fn() };
+    expect(
+      await buildRegistrationTransaction({ client, owner: OWNER, address: shieldedAddress }),
+    ).toBeUndefined();
+    expect(client.getLatestBlockhash).not.toHaveBeenCalled();
+  });
+
+  it("updates differing keys with discriminator 2 and no system account", async () => {
```
ananas-block, line 249: This test takes the accounts and roles from the Rust fixture but substitutes locally computed `data` bytes, so the byte-level half of the assertion compares the code path against itself. What the name claims, discriminator `2` and no system account, is genuinely checked, and the `register` test does cross-check bytes against the fixture. Flagging it so nobody later reads this as byte-level parity coverage for `updateKeys`.

program-libs/user-registry-interface/tests/typescript_wire.rs
```diff
@@ -0,0 +1,85 @@
...
+                u8::from(account.is_signer)
+            )
+        })
+        .collect::<Vec<_>>()
+        .join(" ");
+    format!("{name} {} {hex} {accounts}\n", instruction.program_id)
+}
+
+#[test]
+fn typescript_registry_wire_matches_rust() {
```
ananas-block: Two naming notes:

1. We avoid "wire" in names and comments in this repo; `typescript_fixture.rs` / `registry_fixture_matches_rust` would work.
2. The current name overclaims. The assertion is fixture-vs-Rust, not TypeScript-vs-Rust; the TypeScript half lives in `sdk-libs/ts/test/user-registry.test.ts`. Naming it for what it asserts makes the two-sided arrangement easier to follow.

program-libs/user-registry-interface/tests/typescript_fixture.rs
```diff
@@ -0,0 +1,85 @@
...
+    expected.push_str(&instruction_line(
+        "disableMerging",
+        set_merging_enabled(record, owner, false),
+    ));
+
+    let fixture = concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/../../sdk-libs/ts/fixtures/user-registry.txt"
+    );
+    if std::env::var_os("UPDATE_USER_REGISTRY_FIXTURES").is_some() {
```
ananas-block, line 81: `UPDATE_USER_REGISTRY_FIXTURES` is the only `UPDATE_*` regeneration hook in the repo and is documented nowhere. One line in `sdk-libs/ts/README.md` or in the justfile recipe that runs this test would make it discoverable to the next person who changes `RegisterData`.

sdk-libs/ts/src/interface/pda/index.ts
```diff
@@ -11,11 +11,23 @@ import {
...
 const addressEncoder = getAddressEncoder();
 
+export function userRecordPda(owner: Address): Promise<ProgramDerivedAddress> {
+  return getProgramDerivedAddress({
+    programAddress: USER_REGISTRY_PROGRAM_ID,
+    seeds: [encoder.encode("zolana/registry/v0"), addressEncoder.encode(owner)],
+  });
+}
+
+export async function userRecordAddress(owner: Address): Promise<Address> {
```
ananas-block, line 27: With `userRecordPda` and `userRecordAddress` here, plus the shadowing local `userRecordAddress` (different return type) and `internalUserRecordAddress` / `internalUserRecordPda` in `wallet/registry.ts`, there are now four spellings of "derive the user record address". Not introduced wholesale by this PR, but it adds two of them. Worth collapsing the wallet-side wrappers onto the new canonical pair.

ananas-block:
`sdk-libs/ts/src/wallet/internal.ts:56` - `concat` is dead after this PR. `registry.ts` was its only caller and this change drops the import; nothing else in `src/` or the tests references it. (The other `concat`s are separate definitions in `transaction/internal.ts` and `interface/external-data-hash.ts`.) `oxlint` will not flag it because it is exported. Please delete it.

Posting as a PR comment rather than inline because the file is not in the diff.

ananas-block:
Pre-existing, but this PR makes it visible. `programs/user-registry/src/instructions/update_keys.rs` does `state.owner_p256 = data.owner_p256` unconditionally, so if a record holds a P-256 owner, an ed25519 key rotation through `buildRegistrationTransaction` clears it. The old TypeScript behaved identically, so nothing regresses. Now that `publishedKeysMatch` is the only place left that looks at `ownerP256`, it would be good to know whether clearing is intended.

Posting as a PR comment rather than inline because the file is not in the diff. FYI only, not a request to change anything in this PR.

tilo-14:
1. **The goal is to prevent accidentally clearing an existing P-256 signing key.**

2. **P-256 can exist alongside the Solana Ed25519 owner, with different roles.** The Ed25519 owner controls the registry record and signs updates; the optional P-256 key identifies the user's shielded signing key.

In response to [the P-256 rotation comment](https://github.com/helius-labs/zolana/pull/317#issuecomment-5682950781), the proposed SDK behavior is to register Ed25519 users normally, reject an ordinary registration update that would clear an existing P-256 key, and require an explicit action to switch the record to Ed25519.

This is the proposed follow-up behavior, not a change implemented in this PR. Full P-256 registration and rotation support, including the required proof of possession, is separate work.

PR #319, opened 2026-09-15
fix: reserve nullifier capacity for all UTXO leaves

ananas-block:
## Summary

Spending UTXOs from tree A into tree B can exhaust A's nullifier space while leaving free UTXO leaves. The previous gate compared the two trees' remaining capacities, allowing A to accept deposits whose nullifiers could never be inserted.

Allow dummy and address inputs only when each input tree has enough unreserved nullifier capacity for its UTXO tree's total capacity plus this transaction's inputs assigned to that tree. This prevents a batch from crossing into the reserve. Deposits remain enabled.

Expose the reserve headroom through `dummy_input_headroom(&self)`; transact, merge, and the custom-ring SDK compare it with their input count. Update both protocol-spec references and invariant coverage. Production code changes are limited to four files (+20/-23 lines).

## Validation

- 69 tree tests passed, including reserve equality, batch crossing, invalid capacity, and deposit-independent gating.
- Deposit regression passed with dummy inputs disabled.
- Merge capacity contract check passed.
- Proof-backed transact capacity regression passed.
- Shielded-pool SBF build and all-target checks for zolana-tree, shielded-pool-program, and custom-ring-sdk passed.
- Staged and unstaged diff checks passed.


sergeytimoshin (dismissed):

sdk-tests/dynamic-swap/test/tests/escrow_flow.rs
```diff
@@ -391,9 +393,22 @@ fn create_pair_escrow_and_settle() -> Result<()> {
...
+            // tag. The settler below recovers the order note from a single scan
+            // of that tag and fails outright if the indexer has not caught up.
+            wait_for_indexed_transaction(
+                env.client.indexer(),
+                escrow_owner
+                    .shielded_address()?
+                    .confidential_view_tag()
+                    .map_err(|e| anyhow!("escrow authority view tag: {e:?}"))?,
+                signature,
+            );
```
ananas-block: remove

sdk-tests/dynamic-swap/test/tests/escrow_refund.rs
```diff
@@ -407,9 +409,22 @@ fn create_escrow_underwater_then_refund() -> Result<()> {
...
+        // settler below recovers the order note from a single scan of that tag
+        // and fails outright if the indexer has not caught up yet.
+        wait_for_indexed_transaction(
+            env.client.indexer(),
+            escrow_owner
+                .shielded_address()?
+                .confidential_view_tag()
+                .map_err(|e| anyhow!("escrow authority view tag: {e:?}"))?,
+            signature,
+        );
```
ananas-block: remove

sergeytimoshin (approved):

PR #323, opened 2026-09-21
feat: use poseidon2 for nullifier tree

sergeytimoshin:
Nullifier tree hash: Poseidon → Poseidon2 (BN254, t=2, 6 full / 50 partial rounds). 

One 2-to-1 hash goes from 241 to 186 constraints: transfer 2×3 −8%, merge 8×1 −10%, batch address append −21%.

Open before mainnet: rounds are the gnark-crypto default 6/50, the Poseidon2 paper recommends 8/56.


ananas-block:
Thanks! Interesting experiment, wrong timing and poseidon1 security is better than poseidon2.

